PR #4, opened 2016-04-15
Fix IP-Messaging snippets and meta.json

EliuX:
- Change every snippet´s name to match its correspondent folder
- Add the meta.json  for using the snippets in the Cms as a Code sample
- Check functioning of server side snippets.


mcelicalderon:
LGTM

PR #7, opened 2016-04-19
snippets for video api

llSourcell:
I went ahead and copy-pasted the video snippets from prod to this repo


video/users/client-initialization/example.js
```diff
@@ -0,0 +1,11 @@
+// Create an AccessManager which can manage the expiration and lifecycle of
```
mcelicalderon: we should try to keep the file names the same as the directory, almost sure that is a requirement of the CMS

mcelicalderon:
@llSourcell You should just check that the file names that hold the snippets have the same name of the directory. Also, I think we already added some snippets on the video folder, so you might want to check that before merging. Other than that, it looks good to me

llSourcell:
Thanks! Merging

PR #10, opened 2016-04-19
Includes send-messages snippets

ronualdo:
Includes snippets used in https://www.twilio.com/docs/api/rest/sending-messages

### DIFF
--- a/rest/messages/send-messages/send-messages.cs
+++ b/rest/messages/send-messages/send-messages.cs
@@ -1,0 +1,21 @@
+// Download the twilio-csharp library from twilio.com/docs/csharp/install 
+using System; 
+using Twilio; 
+class Example 
+{ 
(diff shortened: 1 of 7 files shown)


mcelicalderon:
@ronualdo You are missing the output folder that is located in the markdown repository. Also you need to add a meta.json file in each example's directory. Take a look at this as a reference https://github.com/TwilioDevEd/api-snippets/blob/master/ip-messaging/channels/accept-invite/meta.json

the `type` field should be `server` for server side snippets, and `mobile` for client side snippets such as .js .swift and .m files. You can get the snippet title from the markdown file, it will look something like this

```
<div class="md-code-sample"
   id="register-notification-example"
   data-title="Store Registration"
   data-href="guides/ip-messaging/pushiOS/remotenotifications"
   data-pref="mobile">
</div>
```

PR #11, opened 2016-04-20
Send messages copilot snippets

ronualdo:


mcelicalderon:
Same as in #10 , missing output folder and meta.json files. Also, it looks like this branch's base is in #10 's brach, but that shouldn't be a problem when merging.

mcelicalderon:
@ronualdo This shouldn't go inside ip-messaging. The route should be the same as in the markdown repo. In this case, it should go like this:

`api-snippets/rest/messages/send-messages-copilot/`

mcelicalderon:
@ronualdo LGTM :shipit:

PR #12, opened 2016-04-20
Add rest api snippets from old examples

EliuX:


EliuX:
When importing every snippet in a page of the **CMS** we should go to the correspondent code in  `rest` folder and add it a more verbose title

mcelicalderon:
Too many files to check them all, but the structure seems to be correct. We just need to update the title in the meta.json files when we use any snippet. LGTM :shipit:

PR #13, opened 2016-04-20
Instance post example

ronualdo:
Snippets for the REST API: Accounts

### DIFF
--- a/rest/accounts/instance-post-example-3/meta.json
+++ b/rest/accounts/instance-post-example-3/meta.json
@@ -1,0 +1,4 @@
+{
+  "title": "Permanently Close a Subaccount",
+  "type": "server
+}
--- a/rest/accounts/instance-get-example-1/instance-get-example-1.cs
+++ b/rest/accounts/instance-get-example-1/instance-get-example-1.cs
@@ -1,0 +1,17 @@
+// Download the twilio-csharp library from twilio.com/docs/csharp/install
+using System;
+using Twilio;
+class Example 
+{
+  static void Main(string[] args) 
+  {
+    // Find your Account Sid and Auth Token at twilio.com/user/account
+    string AccountSid = "ACba8bc05eacf94afdae398e642c9cc32d";
+    string AuthToken = "{{ auth_token }}";
+    var twilio = new TwilioRestClient(AccountSid, AuthToken);
+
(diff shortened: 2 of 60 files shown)


ip-messaging/rest/accounts/instance-get-example-1/meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "title": "Get account details",
```
mcelicalderon: We should use title case in the snippet titles, I'll check all the existing ones in master to see if they are using this case. But let's fix this from now on

ip-messaging/rest/accounts/list-get-example-1/meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "title": "List all accounts",
```
mcelicalderon: Same here, title case

ip-messaging/rest/accounts/list-get-example-2/meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "title": "List all active accounts",
```
mcelicalderon: Title case, just marking them all so you can see all the commetns gone after the next commits

ip-messaging/rest/accounts/instance-post-example-1/meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "title": "Suspend a subaccount",
```
mcelicalderon: Title case

ip-messaging/rest/accounts/instance-post-example-2/meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "title": "Re-activate a suspended subaccount",
```
mcelicalderon: title case

mcelicalderon:
@ronualdo the same as in #11 , path should be outside `ip-messaging`

rest/accounts/instance-post-example-3/meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "title": "Permanently close a subaccount",
```
mcelicalderon: Title case

PR #19, opened 2016-04-22
Rest resources sip connect

EliuX:
Needed commit by others migrators


EliuX:
Check it @mcelicalderon

PR #53, opened 2016-05-22
Added payfon add-on snippets.

jarodreyes:
Added snippets for the payfone add-on.

@brentschooley Can you take a look at the C# example, since I am having Mono problems today.

### DIFF
--- a/lookups/lookup-get-addon-payfone-example-1/lookup-get-addon-payfone-1.cs
+++ b/lookups/lookup-get-addon-payfone-example-1/lookup-get-addon-payfone-1.cs
@@ -1,6 +1,8 @@
 using Newtonsoft.Json;
 using RestSharp;
 using RestSharp.Authenticators;
+using Newtonsoft.Json.Serialization;
+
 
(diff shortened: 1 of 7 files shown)


brentschooley:
I'd add these `using` statements to be clear about the dependency:

```csharp```
using RestSharp;
using RestSharp.Authenticators;
```

Also, this deserialization code doesn't work by default out of the box for me with the latest version of RestSharp:

```
Console.WriteLine(response.add_ons.results.payfone_tcpa_compliance.result.Description);
Console.WriteLine(response.add_ons.results.payfone_tcpa_compliance.result.Response.NumberMatch);
```

Seems we need to either create a class for the results or parse JSON here.

jarodreyes:
@brentschooley sorry bud, any chance you can commit that change to the snippets? JSON parse is fine. I am afraid I am out of cycles for this specific deliverable. Would appreciate the help.

devinrader:
@jarodreyes pushed changes to the branch that should let the C# example work properly.

PR #66, opened 2016-06-07
Add support to override test session config

mcelicalderon:


tools/snippet-testing/model/test_session_model.rb
```diff
@@ -1,4 +1,28 @@
...
+    def initialize(meta_yaml_path="test.yaml", parent_model = nil)
+      yaml_exists = File.file?(meta_yaml_path)
+      @source_folder = File.dirname(meta_yaml_path)
+      @root_source_folder = parent_model.nil? ? Dir.pwd : parent_model.root_source_folder
+      @relative_folder =@source_folder.sub(root_source_folder,"")
+      @testable = parent_model.nil? ? true : parent_model.testable
+      @root_output_folder = "testable_snippets"
+      if yaml_exists
+       yaml_object = YAML.load(File.read(meta_yaml_path))
+       @testable = yaml_object.fetch("testable") { @testable }
```
mcelicalderon: We should use the second parameter in fetch

tools/snippet-testing/model/test_session_model.rb
```diff
@@ -1,4 +1,28 @@
...
+       @root_output_folder = yaml_object.fetch("output") { parent_model.root_output_folder }
```
mcelicalderon: same here

tools/snippet-testing/snippet_tester.rb
```diff
@@ -7,19 +7,19 @@
   .each { |file| require File.expand_path(file) }
 
 class SnippetTester
-  OUTPUT_PATH = 'testable_snippets'.freeze
+  output_folder = 'testable_snippets'.freeze
```
mcelicalderon: Constant name should be `OUTPUT_FOLDER`

tools/snippet-testing/model/test_session_model.rb
```diff
@@ -1,4 +1,28 @@
...
+
+    alias testable? testable
+
+    def initialize(meta_yaml_path="test.yaml", parent_model = nil)
+      yaml_exists = File.file?(meta_yaml_path)
+      @source_folder = File.dirname(meta_yaml_path)
+      @root_source_folder = parent_model.nil? ? Dir.pwd : parent_model.root_source_folder
+      @relative_folder =@source_folder.sub(root_source_folder,"")
+      @testable = parent_model.nil? ? true : parent_model.testable
+      @root_output_folder = "testable_snippets"
```
mcelicalderon: `testable_snippets` name should be defined only once. This is done in `tools/snippet-testing/snippet_tester.rb`

mcelicalderon:
Right now, if I mark a directory as `testable: false` and then mark a specific snippet within that directory as `"test": "true"` in the meta.json file, that snippet won't be tested. meta.json configuration should override everything else.

PR #67, opened 2016-06-07
Replace mustache placeholders

mcelicalderon:
For snippet testing to work, user data placeholders should use a format that resembles the original value. 

```
Account SID: ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
Call SID: CAXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,24 @@
-# api-snippets
+# Code snippets for the Twilio API documentation
 [![Build Status](https://travis-ci.org/TwilioDevEd/api-snippets.svg?branch=master)](https://travis-ci.org/TwilioDevEd/api-snippets)
 
-Code snippets for the Twilio API Documentation
+## Guidelines
+
+1. Snippets should use placeholders for user information in a format that resembles
+   the original value. For example:
+   ```
+   Account SID: ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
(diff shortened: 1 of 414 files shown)


mcelicalderon:
@kwhinnery @jarodreyes @dprothero @atbaker Do you think we should follow some guidelines for code snippets, like the one I specified in this PR's readme file?

In this PR I just replaced every `{{ account_sid }}` with `ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX`

Please let me know what you think.

dprothero:
I think the idea behind the mustache placeholders was so we could eventually dynamically replace them with the logged-in user's credentials. However, as long as we use a consistent value -- and it's not a value likely to occur in any other context -- then I think we should be fine.

mcelicalderon:
@dprothero Yes, consistency is the most important factor here. In the past we did had a conversation about using mustache or a format that resembles the actual value. In the end, we decided to go with the second option. But, if anyone thinks mustache is better for any reason I can adapt to that, I just need to make it consistent.

mcelicalderon:
I will merge this as we can change it later if that decision was to be made. For now, we will have consistency as many placeholders already use the `ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX` format.

PR #73, opened 2016-06-10
fix end_date

meganspeir:
Fix 2nd start_date to end_date.


atbaker:
Wooo thanks @meganspeir!

PR #75, opened 2016-06-10
Find and replace python 2 print for python 3

eddiezane:
```
ack -l 'print .*' --python | xargs perl -pi -E 's/print (.*)/print\($1\)/g'
```


atbaker:
Beautiful - thanks @eddiezane!

Btw @mcelicalderon & @EliuX - our preference going forward for Python code snippets is to use `print` as a function, like this:

**Good:** `print('hello world')`
**Bad:** `print 'hello world'`

Why? Because `print('hello world')` is compatible with Python 2 and Python 3, but `print 'hello world'` is only compatible with Python 2, which is on its way out.

We will take care of changing the existing inline code samples to use `print()` instead of `print`.

PR #77, opened 2016-06-10
Fix IPM send-messages node example

eddiezane:


atbaker:
BOOM! Thanks @eddiezane!

PR #85, opened 2016-06-15
Replace all mustache placeholders

mcelicalderon:


mcelicalderon:
@kwhinnery @dprothero @atbaker @jarodreyes I'll be replacing the placeholders to match the new format. This is necessary for taskrouter snippets to pass the tests.

kwhinnery:
lgtm - let's make sure we spot check a few dozen pages in the API reference across all the different products to make sure things are rendering properly.

mcelicalderon:
Sure @kwhinnery It usually takes a while to update them all, but they eventually render correctly. @atbaker please take a look at the CMS logs to see if an error occurred after this merge.

PR #90, opened 2016-07-05
Add next-get PHP snippets

mcelicalderon:


kwhinnery:
Couple bits:

- Can we go with `example.5.x.php` rather than `example.5.php` since the `5.x` string will be used in the UI to switch between versions
- We would also need to rename existing snippets to `example.4.x.php` to show the difference
- is this `getContext` thing actually necessary? I would think that `$client->api->accounts('ACXXX')->fetch();` would be a thing?

mcelicalderon:
Hey @kwhinnery ! I sent an email regarding the first two observations :)   For the third one, accounts is actually the only case where you can't do that. Probably something that can be changed in the library.

PR #94, opened 2016-07-11
Fix workspace fetch example

jwitz10:


mcelicalderon:
Thanks for the fix @jwitz10 ! I can merge that if those are all the changes you will introduce in this PR.

jwitz10:
@mcelicalderon yup this is it, pls merge away!

PR #95, opened 2016-07-12
Removing alternative csharp example; does not work

wanjunsli:


mcelicalderon:
Hey @wanjunsli why doesn't it work? There is actually an overloaded method [here](https://github.com/twilio/twilio-csharp/blob/master/src/Twilio.TaskRouter/Tasks.cs#L49-L53) that should convert the dictionary to a JSON string. What version of the library are you using?

wanjunsli:
Hey @mcelicalderon!

There is a bug in the C# helper lib on the FromDictionaryToJson method, there's a PR out to fix it (https://github.com/twilio/twilio-csharp/pull/244), but it's yet to be merged. This is taken care of in yoyodeyne.

mcelicalderon:
I see, thanks!

PR #96, opened 2016-07-12
Add new taskrouter snippets for twilio-php v5

EliuX:


pricing/test.yml
```diff
@@ -1 +0,0 @@
-testable: true
```
mcelicalderon: @EliuX this shouldn't be commited in this PR.

rest/taskrouter/activities/instance/delete/example-1/example-1.5.x.php
```diff
@@ -0,0 +1,19 @@
+<?php
+// Get the PHP helper library from twilio.com/docs/php/install
+require_once('/path/to/twilio-php/Services/Twilio.php'); // Loads the library
```
mcelicalderon: As PHPCS will suggest (I think that the PHP interpreter will also give you a warning) `require_once` is not a function you should not use parenthesis. This is an error found in the v4 snippets. Already fixed in my v5 snippets.

rest/taskrouter/activities/instance/delete/example-1/example-1.5.x.php
```diff
@@ -0,0 +1,19 @@
...
+
+use Twilio\Rest\Client;
```
mcelicalderon: I saw you changed this in all the snippets. Any reason why you decided not to use the global namespace for every snippet? This works only because we are not using namespaces in this context, might fail if the snippet is copied into a file that is actually using a namespace

EliuX: Based on the code used in the tutorial code of task-router-laravel this is how i should use the import, and PHPStorm formats it this way. It do works with a file with a namespace, e.g, https://github.com/TwilioDevEd/task-router-laravel/blob/master/app/TaskRouter/WorkspaceFacade.php.

PR #107, opened 2016-07-26
Sync C# samples

dprothero:


atbaker:
👍  from me - will wait on @dprothero to merge

PR #108, opened 2016-07-27
Sync PHP samples

dprothero:


kwhinnery:
Can we use `json_encode` on an actual object rather than manually creating the string? I feel like that's more likely what a developer would do.

dprothero:
@kwhinnery Done. Look good?

kwhinnery:
LGTM

PR #109, opened 2016-07-27
Added all missing sync ruby snippets, and updated a few.

jarodreyes:


sync/rest/lists/query-list/query-list.rb
```diff
@@ -0,0 +1,18 @@
...
+auth_token = 'your_auth_token'
+client = Twilio::REST::Client.new(account_sid, auth_token)
+
+# Retreive the service
+service = client.preview.sync.services('ISXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX')
+
+# Query Sync List
+items = service.sync_lists('MyCollection').sync_list_items.stream(
+	from: '10',
+	directions: 'forward'
```
dprothero: @jarodreyes I think this should be 'direction', not 'directions'?

dprothero:
LGTM

PR #115, opened 2016-07-30
Sms snippets

jarodreyes:
Added snippets for a bunch of basic SMS activities.


jarodreyes:
Well it certainly seems like copy/paste from the quickstarts is not going to work... boo. Will take a look at writing from scratch this weekend.

mcelicalderon:
@jarodreyes Just FWI build is failing because of the TwiML snippets. Testing is not currently supported for snippets that use and run a web framework like Flask and Sinatra as an actual request would have to be made to the server that runs upon snippet execution. But you can mark a snippet not to be tested in the meta.json file or a whole directory by adding a `test.yml` file whose content is `testable: false`. Also, there might be a better place for TwiML snippets as they are not actual `rest/` snippets. Maybe something similar to what we have in the `voice/` dir would be a better approach.

jarodreyes:
@mcelicalderon Ready for you bud! Just need the latest php 5.x snippets.

mcelicalderon:
Great, I'll create those later today 👍

mcelicalderon:
@jarodreyes done. Just one thing. A snippet to send sms already exists [here](https://github.com/TwilioDevEd/api-snippets/tree/master/rest/messages/send-message). The one you are creating `send-sms` is almost the same. The only difference is that in the new one you don't provide a media url for the created sms.

jarodreyes:
@mcelicalderon Correct we need a basic SMS example without mediaUrl.

mcelicalderon:
Ok then, merging as soon as the build passes. 👍

PR #117, opened 2016-08-01
Add Java next-gen snippets in the REST/ directory

mcelicalderon:


mcelicalderon:
@jingming Could you please take a look at this PR to see if I'm in the right path? Any suggestion or correction is welcome as I have to create tons of snippets and I would like to use your feedback if any for the new ones. Thanks!

rest/accounts/instance-get-example-1/instance-get-example-1.7.x.java
```diff
@@ -0,0 +1,20 @@
...
+  // Find your Account Sid and Token at twilio.com/user/account
+  public static final String ACCOUNT_SID = "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX";
+  public static final String AUTH_TOKEN = "your_auth_token";
+
+  public static void main(String[] args) {
+    Twilio.init(ACCOUNT_SID, AUTH_TOKEN);
+
+    // Get an object from its sid. If you do not have a sid,
+    // check out the list resource examples on this page
+    Account account = new AccountFetcher(ACCOUNT_SID).execute();
```
jingming: For 2010 resources, we also provide a convenience constructor so that you don't have to pass in the account sid. We will then pull the account sid off of the client. 

`new AccountFetcher().execute();`

mcelicalderon: I think the idea behind this snippet is to show how to fetch an account by SID, not necessarily the authenticated account in the client, maybe a sub account. But I will change it so it doesn't use the same `ACCOUNT_SID `

PR #126, opened 2016-08-17
Add Mustache template placeholders

mcelicalderon:


mcelicalderon:
Hey @kwhinnery we are not going to go through with this, right? Should we close this PR?

acamino:
@mcelicalderon does it makes sense to have this pull request still opened? Seems to me that would be good if we close it.

PR #128, opened 2016-08-18
Add Mustache templates to snippets

mcelicalderon:


mcelicalderon:
@atbaker @kwhinnery @crtr0 Creating this first template snippet took me like 25 minutes, just one language. But this one sums up 4 snippets into one. Took a bit longer than the next ones probably as I had to look a bit into the library to check for method signatures.

rest/making-calls/template/example-1.4.x.php
```diff
@@ -0,0 +1,26 @@
...
+$call = $client->account->calls->create(
+    {{#fromPhoneNumber}}"{{fromPhoneNumber}}",{{/fromPhoneNumber}}
+    {{#toPhoneNumber}}"{{toPhoneNumber}}",{{/toPhoneNumber}}{{#toVoiceClient}}"{{toVoiceClient}}",{{/toVoiceClient}}
+    {{#callUrl}}"{{callUrl}}",{{/callUrl}}
+    array(
+        {{#sendDigitsVoice}}"SendDigits" => "{{#sendDigitsVoice}}",{{/sendDigitsVoice}}
+        {{#callMethod}}"Method" => "{{#callMethod}}",{{/callMethod}}
+        {{#callStatusCallback}}"StatusCallback" => "{{callStatusCallback}}",{{/callStatusCallback}}
+        {{#callStatusCallbackMethod}}"StatusCallbackMethod" => "{{callStatusCallbackMethod}}",{{/callStatusCallbackMethod}}
+        {{#callStatusCallbackEvent}}
```
mcelicalderon: For list of parameters we might need to set an extra value like `callStatusCallbackEvent` if any value is selected and then assign the actual list of values to `callStatusCallbackEvents` for example.

rest/making-calls/template/example-1.4.x.php
```diff
@@ -0,0 +1,26 @@
+<?php
+// Get the PHP helper library from twilio.com/docs/php/install
+require_once('/path/to/twilio-php/Services/Twilio.php'); // Loads the library
+
+// Your Account Sid and Auth Token from twilio.com/user/account
+{{#accountSid}}$sid = "{{accountSid}}";{{/accountSid}}
+{{#authToken}}$token = "{{authToken}}";{{/authToken}}
```
kwhinnery: These would be required, so we wouldn't need conditionals

rest/making-calls/template/example-1.4.x.php
```diff
@@ -0,0 +1,26 @@
...
+$client = new Services_Twilio($sid, $token);
+
+$call = $client->account->calls->create(
+    {{#fromPhoneNumber}}"{{fromPhoneNumber}}",{{/fromPhoneNumber}}
+    {{#toPhoneNumber}}"{{toPhoneNumber}}",{{/toPhoneNumber}}{{#toVoiceClient}}"{{toVoiceClient}}",{{/toVoiceClient}}
+    {{#callUrl}}"{{callUrl}}",{{/callUrl}}
```
kwhinnery: for required parameters, again, wouldn't need to be conditional.

rest/making-calls/template/example-1.6.x.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.http.NameValuePair;
+import org.apache.http.message.BasicNameValuePair;
+
+public class Example {
+
+  // Find your Account Sid and Token at twilio.com/user/account
+  public static final String ACCOUNT_SID = "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX";
+  public static final String AUTH_TOKEN = "your_auth_token";
```
kwhinnery: need to templatize

rest/making-calls/template/example-1.4.x.php
mcelicalderon: Most libraries (if not all) support using an (apiKey, apiSecret) to authenticate. But we can probably separate those snippets into just one that allows you to do that?

mcelicalderon:
@atbaker @kwhinnery @crtr0 this PR has a template for every language/library version. This would be the `Create` action for the `Call` resource. Every possible parameter for this action is considered.

mcelicalderon:
Things to take into account:

- Trailing commas in parameter lists. The last argument of a list or array in some languages can't have a trailing comma. How will that be handled when replacing placeholders?
- For java and C# we also have conditional `import` and `using` statements.
- At least in this PR boolean parameters and "true" or "false" strings all have the same casing. Might be a problem for other values.
- The same for `HTTP Methods`. We can replace `GET`, `POST` seamlessly in this PR. Casing might  be an issue for other values as they are not all strings but static variable names (java, C#).
- An additional value needs to be set for `StatusCallbackEvent` as creating an array or list needs two additional lines of code.

kwhinnery:
think we're going to handle this differently, thanks for the exploration, however.

PR #136, opened 2016-09-07
Retrieve modify message history

acamino:
This one completes the snippets for PHP and Node.js.


rest/message/list-get-example-2/list-get-example-2.5.x.php
```diff
@@ -1,15 +1,18 @@
 <?php
 // Get the PHP helper library from twilio.com/docs/php/install
-require_once '/path/to/vendor/autoload.php'; // Loads the library
+require_once 'vendor/autoload.php'; // Loads the library
```
acamino: I'm sure we'd like to preserve `/path/to/vendor/autoload.php`.

PR #138, opened 2016-09-07
Add PHP snippets to respond-incoming-phone-call

mosampaio:


mosampaio:
Used by https://www.twilio.com/docs/admin/pages/1566/

kwhinnery:
looks good - no need to implement both old and new versions at this juncture, 5.x alone is fine for now.

PR #147, opened 2016-09-19
Update Examples to Use the Messages Resource

acamino:
### TODO
- [x] Squash the commits.

### DIFF
--- a/rest/test-credentials/test-sms-messages-example-1/test-sms-messages-example-1.4.x.php
+++ b/rest/test-credentials/test-sms-messages-example-1/test-sms-messages-example-1.4.x.php
@@ -1,11 +1,19 @@
 <?php
+
 // Get the PHP helper library from twilio.com/docs/php/install
-require_once('/path/to/twilio-php/Services/Twilio.php'); // Loads the library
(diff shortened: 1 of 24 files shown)


rest/test-credentials/test-sms-messages-example-1/test-sms-messages-example-1.4.x.php
```diff
@@ -7,5 +8,10 @@
 $token = "your_auth_token";
```
acamino: Should we have some naming convention for the _account SID_ and the _authentication token_?
It will be more consistent with the other examples if we use something like:

```php
$accountSid = "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"; 
$authToken = "your_auth_token"; 
$client = new Services_Twilio($accountSid, $authToken);
```

What do you think?

mcelicalderon: 👍

acamino: Thanks!

mcelicalderon:
@acamino LGTM :shipit:

PR #173, opened 2016-10-12
Fix include instructions for twilio-video.js

markandrus:
@aaalaniz

### DIFF
--- a/video/getting-started/create-video-client/create-video-client.js
+++ b/video/getting-started/create-video-client/create-video-client.js
@@ -1,10 +1,10 @@
 // If you are requiring twilio-video from CommonJS,
 //
-//     const Video = Twilio.Video;
+//     const Video = require('twilio-video');
 //
 //  If you are including twilio-video.js from a <script> tag,
 //
-//     const Video = require('twilio-video');
+//     const Video = Twilio.Video;
 //
 
 const client = new Video.Client(accessToken);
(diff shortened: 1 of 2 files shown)


video/getting-started/create-video-client/create-video-client.js
```diff
@@ -1,10 +1,10 @@
 // If you are requiring twilio-video from CommonJS,
```
manjeshbhargav: @markandrus I would rewrite this as:
```javascript
// If you are using CommonJS module pattern, then you can include the SDK like this:
// const Video = require('twilio-video');
```

video/getting-started/setup-local-media/setup-local-media.js
```diff
@@ -1,10 +1,10 @@
 // If you are requiring twilio-video from CommonJS,
```
manjeshbhargav: @markandrus Same as above.

manjeshbhargav:
@markandrus Looks good. I'll leave it up to you to decide whether you want to include my suggestion about wording the comment for `require()` differently.

video/getting-started/create-video-client/create-video-client.js
markandrus: I think "requiring" is a little more direct than "use", and makes sense in the context of CommonJS.

markandrus:
Thanks for the feedback @manjeshbhargav

PR #177, opened 2016-10-14
Do not test snippets unless library version is specified in file name

mcelicalderon:
Snippet file name must specify a Twilio library version. Disable testing legacy version by default. Warning will be printed in the error log for files not marked with an specific version and the build will fail.


mcelicalderon:
@acamino please take a look. In order to merge this, we would need to rename every snippet to the correct library version.

PR #189, opened 2016-10-27
Add remaining Ruby snippets

acamino:
By default a folder that does not contain a file `test.yml` is not testable. But you have to make sure the code will work in your machine.

The testable snippets in the following list are: _pricing_ and _security_.
### TODO
- [x] client
- [x] guides
- [x] insights
- [x] ip-messaging
- [x] lookups
- [x] monitor
- [x] notifications
- [x] **pricing**
- [x] **security**
- [x] stun-turn
- [x] sync
- [x] video
- [x] voice
- [x] wireless

/cc @joliveros, @juancarlospaco, @hortega


acamino:
There is no need to add a Sinatra dependency -- the snippets that has code that includes dependencies like Sinatra are not testable -- If you ran some example that includes something like Sinatra, Flask an so on it will run forever unless you explicitly send to it a SIGINT which will interrupt the application, which makes the snippet untestable.

There are a few rules that determines if a folder is testable:

1. The folder must contain a _test.yml_ file which contains a propery `testable: true`.
1. The folder contain a file `meta.json` which specify if the examples in that folder are testable.
1. **If none of the previous files are present the folder is non testable**.

client/capability-token-2way/capability-token.5.x.rb
```diff
@@ -0,0 +1,19 @@
...
+require 'sinatra'
+
+get '/token' do
+  account_sid = 'ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'
+  auth_token = 'your_auth_token'
+  capability = Twilio::JWT::Capability.new(account_sid, auth_token)
+
+  # Create an application sid at
+  # twilio.com/console/phone-numbers/dev-tools/twiml-apps and use it here
+  capability.allow_client_outgoing("APXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX")
```
joliveros: I would avoid using double quotes and single quotes unless, as Agustin mentioned, we are doing string interpolation.

guides/voice/gather-dtmf-tones-guide/gather-example-step-0/example.5.x.rb
```diff
@@ -0,0 +1,13 @@
...
+require 'rubygems' # This line not needed for ruby > 1.8
+require 'sinatra'
+require 'twilio-ruby'
+
+post '/voice' do
+  Twilio::TwiML::Response.new do |r|
+    r.gather numDigits: 1 do |g|
+      g.say 'For sales, press 1. For support, press 2.'
+    end
+    r.redirect('/voice')
```
joliveros: I would add some space between blocks of code.

ip-messaging/rest/channels/create-channels/create-channels.5.x.rb
```diff
@@ -0,0 +1,15 @@
+require 'http'
+require 'twilio-ruby'
+
+# Initialize the client
+account_sid = 'ACCOUNT_SID'
+auth_token = 'AUTH_TOKEN'
+ip_messaging_client = Twilio::REST::IpMessagingClient.new(
```
acamino: TL;TR

This one won't work. The _next-gen_ version does not have an `Twilio::REST::IpMessagingClient` -- Keep in mind that the next version only exposes a `Twilio::REST::Client`.

Besides that, there is no `get` method exposed in the public interface of the next-gen Twilio SDK.

You can see that the hierarchy of classes goes like this:

```
ListResource
  InstanceContext
    InstanceResource
```
Where:

| Base class | Defined methods |
|---|---|
| `ListResource` | `list`, `create`|
| `InstanceContext` | `fetch` |
| `ListResource` | `delete`, `update`|

e.g.:

```ruby
client = Twilio::REST::Client(...)

client.calls #=> CallList < ListResource
client.account.calls('some random SID') #=> CallContext < InstanceContext
client.account.calls('some random SID').fetch #=> CallInstance < InstanceResource
```

acamino:
Thank you guys!

joliveros:
and...merging!

PR #190, opened 2016-11-01
Improve Travis Performance

mcelicalderon:


.travis.yml
```diff
@@ -1,14 +1,10 @@
 sudo: required
 
-language: csharp
-
-mono:
-  - latest
+rvm:
+  - 2.2.1
```
acamino: What do you think about removing those extra spaces? Kinda picky but I think that one single space will be enough to provide a proper separation between steps. _Travis documentation uses a single space to separate steps._

acamino:
Definitely it LGTM. 😉
We'll merge this one as soon as we get a green build.

PR #197, opened 2016-11-10
Adds Hold field and Example to Participant Snippits

akjordan:
Kicking off getting hold snippets done. Still todo is write up examples for the next gen libs which support the hold param.


atbaker:
+1 from me

PR #200, opened 2016-11-30
Add request-validation-express/example-1

joliveros:


joliveros:
@acamino this PR is ready for your review. 

Thanks!

guides/request-validation-express/example-1/example-1.3.x.js
```diff
@@ -0,0 +1,35 @@
...
+app.post('/voice', twilio.webhook(), function(req, res) {
+  // Twilio Voice URL - receives incoming calls from Twilio
+  var response = new TwimlResponse();
+
+  response.say(
+    `Thanks for calling!
+     Your phone number is ${req.body.From}. I got your call because of Twilio´s webhook.
+     Goodbye!`
+  );
+
```
acamino: Did you consider setting the content type to `text/xml`? It would be appropriate since we're responding with TwiML.

```javascript
response.set('Content-Type', 'text/xml');
```

guides/request-validation-express/example-1/example-1.3.x.js
```diff
@@ -0,0 +1,35 @@
...
+// Please note that this will not work unless you set the TWILIO_AUTH_TOKEN
+// environment variable.
+var twilio = require('twilio');
+var app = require('express')();
+var bodyParser = require('body-parser');
+var TwimlResponse = twilio.TwimlResponse;
+
+app.use(bodyParser.urlencoded({extended: false}));
+
+app.post('/voice', twilio.webhook(), function(req, res) {
```
acamino: What do you think about using arrow functions to keep consistency with the use of the ES6 syntax? We're already using template literals, right?

guides/request-validation-express/example-1/example-1.3.x.js
```diff
@@ -0,0 +1,35 @@
+// You can find your Twilio Auth Token here: https://www.twilio.com/console
+// Which can be set at runtime as follows:
+// $ TWILIO_AUTH_TOKEN=XXXXXXXXXXXXXXXXXXX node index.js
+// Please note that this will not work unless you set the TWILIO_AUTH_TOKEN
+// environment variable.
+var twilio = require('twilio');
```
acamino: Did you consider using `const`?

guides/request-validation-express/example-1/meta.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "Custom middleware for Express apps that validates Twilio requests",
```
acamino: Not sure if this title still reflects what we did in the this implementation.

guides/request-validation-express/example-1/example-1.3.x.js
```diff
@@ -0,0 +1,35 @@
...
+var app = require('express')();
+var bodyParser = require('body-parser');
+var TwimlResponse = twilio.TwimlResponse;
+
+app.use(bodyParser.urlencoded({extended: false}));
+
+app.post('/voice', twilio.webhook(), function(req, res) {
+  // Twilio Voice URL - receives incoming calls from Twilio
+  var response = new TwimlResponse();
```
joliveros: I would only use let in the case that we were going to reuse or redefine a variable. In this case response might be a candidate. But I think we are safer with const. Where would you use it in this example?

acamino: Agreed, thanks!

acamino:
I don't want to delay this pull request, that's why the last commit introduces a couple of minor cosmetic changes.

Keep up with the good job man! Thank you!

PR #201, opened 2016-12-01
Add request-validation-sinatra

juancarlospaco:
- Add request-validation-sinatra examples.
- Rubocop clean and tidy.
- Add meta.

### DIFF
--- a/guides/request-validation-sinatra/example-1/example-1.3.x.rb
+++ b/guides/request-validation-sinatra/example-1/example-1.3.x.rb
@@ -1,9 +1,6 @@
 require 'sinatra'
 require 'twilio-ruby'
 require 'rack'
-require 'dotenv'
-
-Dotenv.load
 
 use Rack::TwilioWebhookAuthentication, ENV['TWILIO_AUTH_TOKEN'], '/'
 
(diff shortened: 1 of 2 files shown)


guides/request-validation-sinatra/example-1/example-1.3.x.rb
```diff
@@ -0,0 +1,17 @@
+require 'sinatra'
+require 'twilio-ruby'
+require 'rack'
+require 'dotenv'
```
acamino: I don't think `dotenv` is required in the scope of this guide. When you introduce something `dotenv` you should also introduce a **.env** file which holds the env variables. For the purpose of this guide that would be unnecessary complexity.

What do you think?

guides/request-validation-sinatra/example-1/example-1.3.x.rb
```diff
@@ -0,0 +1,17 @@
+require 'sinatra'
```
acamino: A good approach is to introduce an explanatory commentary, similar to the ones we already introduced to the other code snippets. Did you consider adding a comment like this:

```
// You can find your Twilio Auth Token here: https://www.twilio.com/console
// Set at runtime as follows:
// $ TWILIO_AUTH_TOKEN=XXXXXXXXXXXXXXXXXXX node index.js
//
// This will not work unless you export the TWILIO_AUTH_TOKEN environment variable.
//
```

acamino:
Looks good man. Please address the last comment on blank lines usage and we're ready. LGTM
Please merge it right after the build is green. Thanks!

PR #206, opened 2016-12-02
Add snippets for request-validation-csharp

hortega:


guides/request-validation-csharp/example-1/example-1.4.x.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Configuration;
+using System.Web;
+using System.Net;
+using System.Web.Mvc;
+using Twilio;
+using Twilio.TwiML;
+
+namespace helloworldcsharp
+{
+	public class RequestValidatorFilterAttribute : ActionFilterAttribute
```
acamino: Additionally, I suggest to apply `AttributeUsage` to this class to limit the application scope for this attribute to only methods, e.g.:

```csharp
[AttributeUsage(AttributeTargets.Method)]
public class ValidateTwilioRequestAttribute : ActionFilterAttribute
{
    // ...
```

guides/request-validation-csharp/example-1/example-1.4.x.cs
```diff
@@ -0,0 +1,30 @@
...
+	{
+
+		private static readonly Lazy<RequestValidator> TwilioRequestValidator
+			= new Lazy<RequestValidator>(() => new RequestValidator());
+
+		public override void OnActionExecuting(ActionExecutingContext filterContext)
+		{
+			var twilioAuthToken = ConfigurationManager.AppSettings["TwilioAuthToken"];
```
acamino: Usually is not a good idea to hardcode values in any implementation. This this implementation is coupled to the idea of having a key in the _application settings_ named "TwilioAuthToken".

A more idiomatic approach is to use an overloaded constructor. When no value is provided then we can rely on a fallback value "TwilioAuthToken", otherwise we can use the provided key through the constructor.

```csharp
private readonly string _twilioAuthTokenKey;

public ValidateTwilioRequestAttribute(string twilioAuthTokenKey)
{
    _twilioAuthTokenKey = twilioAuthTokenKey;
}

public ValidateTwilioRequestAttribute()
{
    _twilioAuthTokenKey = "TwilioAuthToken";
}

public override void OnActionExecuting(ActionExecutingContext filterContext)
{
    var twilioAuthToken = ConfigurationManager.AppSettings[_twilioAuthTokenKey];
    // ...

```

guides/request-validation-csharp/example-1/example-1.4.x.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Web.Mvc;
+using Twilio;
+using Twilio.TwiML;
+
+namespace helloworldcsharp
+{
+	public class RequestValidatorFilterAttribute : ActionFilterAttribute
+	{
+
+		private static readonly Lazy<RequestValidator> TwilioRequestValidator
```
acamino: No need declare this field as _static_, please consider to use:

```csharp
private readonly RequestValidator _requestValidator = new RequestValidator();
```

guides/request-validation-csharp/example-1/example-1.4.x.cs
```diff
@@ -0,0 +1,30 @@
...
+		{
+			var twilioAuthToken = ConfigurationManager.AppSettings["TwilioAuthToken"];
+			var context = ((HttpApplication)filterContext.HttpContext.GetService(typeof(HttpApplication))).Context;
+			var isValidRequest = TwilioRequestValidator.Value.IsValidRequest(context, twilioAuthToken);
+
+			if (!isValidRequest)
+			{
+				filterContext.Result = new System.Web.Mvc.HttpStatusCodeResult(HttpStatusCode.Forbidden);
+			}
+			base.OnActionExecuting(filterContext);
```
acamino: A line before `base` would be nice.

guides/request-validation-csharp/example-2/example-2.4.x.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using System.Configuration;
+using System.Web;
+using System.Web.Mvc;
+using Twilio.TwiML;
+using Twilio.TwiML.Mvc;
+
+namespace helloworldcsharp.Controllers
```
acamino: This one should be `HelloWorldCSharp.Controllers`.

guides/request-validation-csharp/example-2/example-2.4.x.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+	public class IncomingController : Controller
```
acamino: This is the perfect opportunity to introduce a [TwilioController](https://github.com/twilio/twilio-csharp/blob/ab0983ae00269746c5dbeb891955ae3edd83796d/src/Twilio.Mvc/TwilioController.cs). What do you think?

guides/request-validation-csharp/example-2/example-2.4.x.cs
```diff
@@ -0,0 +1,41 @@
...
+	{
+		[RequestValidatorFilter]
+		public ActionResult Voice (string From)
```
acamino: No need to use capitalize `From` it would work as expected with `from`. Using camel casing for variables is the preferred approach in C#.

guides/request-validation-csharp/example-2/example-2.4.x.cs
```diff
@@ -0,0 +1,41 @@
...
+		{
+			var response = new TwilioResponse();
+			var message = @"
```
acamino: Is recommended to use `const` when the value is immutable.

```csharp
const string message = @"
    Thanks for calling!
    Your phone number is {0}. I got your call because of Twilio's webhook.
    Goodbye!";
```

guides/request-validation-csharp/example-2/example-2.4.x.cs
```diff
@@ -0,0 +1,41 @@
...
+				Thanks for calling!
+
+    			Your phone number is {0}. I got your call because of Twilio's webhook.
+
+    			Goodbye!
+			";
+			response.Say(string.Format(message, From));
+			response.Hangup();
+
+			return new TwiMLResult(response);
```
acamino: The reason to using `TwilioController` as the base class for this controller is to use the `TwiML` helper, which is just a shortcut to what you did in this line.

```charp
return TwiML(response);
```

guides/request-validation-csharp/example-2/example-2.4.x.cs
```diff
@@ -0,0 +1,41 @@
...
+		}
+
+		[RequestValidatorFilter]
+		public ActionResult Message (string Body)
+		{
+			var response = new TwilioResponse();
+			var message = "Your text to me was {0} characters long. Webhooks are neat :)";
```
acamino: ditto - about `const` usage.

guides/request-validation-csharp/example-3/example-3.4.x.cs
```diff
@@ -0,0 +1,37 @@
...
+	{
+
+		private static readonly Lazy<RequestValidator> TwilioRequestValidator
+			= new Lazy<RequestValidator>(() => new RequestValidator());
+
+		private readonly bool IsTestEnvironment;
+
+		public RequestValidatorFilterAttribute ()
+		{
+			this.IsTestEnvironment = bool.Parse(ConfigurationManager.AppSettings["IsTestEnvironment"]);
```
acamino: I think we can use the approach proposed before to avoid coupling this implementation with the name `IsTestEnvironment `. What do you think?

guides/request-validation-csharp/example-1/example-1.4.x.cs
```diff
@@ -0,0 +1,30 @@
+using System;
```
acamino: We should clean up the `using`s. Is a good practice to keep only the required ones.

guides/request-validation-csharp/example-2/example-2.4.x.cs
```diff
@@ -0,0 +1,41 @@
+using System;
```
acamino: ditto - about clean up your `using`s.

guides/request-validation-csharp/example-1/example-1.4.x.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Configuration;
+using System.Web;
+using System.Net;
+using System.Web.Mvc;
+using Twilio;
+using Twilio.TwiML;
+
+namespace helloworldcsharp
```
hortega: thanks for the analogy, much more clear. I will also change the 'helloworld' bit, does not make sense for this example

guides/request-validation-csharp/example-2/example-2.4.x.cs
```diff
@@ -0,0 +1,41 @@
...
+			return new TwiMLResult(response);
+		}
+
+		[RequestValidatorFilter]
+		public ActionResult Message (string Body)
+		{
+			var response = new TwilioResponse();
+			var message = "Your text to me was {0} characters long. Webhooks are neat :)";
+
+			response.Say(string.Format(message, Body.Length));
```
acamino: I almost miss this one but you can use [interpolated strings](https://msdn.microsoft.com/en-us/library/dn961160.aspx) and remove the `message` variable. What do you think?

```csharp
response.Say($"Your text to me was {body.Length} characters long. Webhooks are neat :)");
```

guides/request-validation-csharp/example-1/example-1.4.x.cs
hortega: It seems like every single IDE defaults to tabs :laughing:

guides/request-validation-csharp/example-2/example-2.4.x.cs
hortega: yes! I was looking for something like that! thanks

guides/request-validation-csharp/example-3/example-3.4.x.cs
hortega: sure, it's an useful pattern

guides/request-validation-csharp/example-1/example-1.4.x.cs
```diff
@@ -0,0 +1,42 @@
+using System;
+using System.Configuration;
+using System.Web;
+using System.Net;
+using System.Web.Mvc;
+using Twilio;
```
acamino: This reference to `Twilio` is not required here. You only use `Twilio.TwiML`. Would you mind removing this line please?

guides/request-validation-csharp/example-1/example-1.4.x.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Net;
+using System.Web.Mvc;
+using Twilio.TwiML;
+
+namespace ValidateRequestExample.Filters
+{
+    [AttributeUsage(AttributeTargets.Method)]
+    public class ValidateTwilioRequestAttribute : ActionFilterAttribute
+    {
+        private readonly RequestValidator TwilioRequestValidator = new RequestValidator();
```
acamino: Usually we prefer using an `_` to prefix the camel cased variable for a private field. This one would be `_twilioRequestValidator` or even better `_requestValidator` - this feels more natural for a C# developer. Could you please update it?

guides/request-validation-csharp/example-1/example-1.4.x.cs
```diff
@@ -0,0 +1,40 @@
...
+        private readonly string TwilioAuthTokenKey;
```
acamino: ditto - about naming conventions for private fields.

guides/request-validation-csharp/example-1/example-1.4.x.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+
+        public override void OnActionExecuting(ActionExecutingContext filterContext)
+        {
+            var twilioAuthToken = ConfigurationManager.AppSettings[TwilioAuthTokenKey];
+            var context = ((HttpApplication)filterContext.HttpContext.GetService(typeof(HttpApplication))).Context;
+            var isValidRequest = TwilioRequestValidator.IsValidRequest(context, twilioAuthToken);
+
+            if(!isValidRequest) {
```
acamino: Kinda picky, but this `{` should be in the next line. 😉

guides/request-validation-csharp/example-2/example-2.4.x.cs
```diff
@@ -0,0 +1,38 @@
+using System.Configuration;
```
acamino: Seems like this class doesn't use `System.Configuration` and `System.Web`.

guides/request-validation-csharp/example-2/example-2.4.x.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            var response = new TwilioResponse();
+            const string message = @"
+		          Thanks for calling!
+  			      Your phone number is {0}. I got your call because of Twilio's webhook.
+    			    Goodbye!";
+            response.Say(string.Format(message, from));
+            response.Hangup();
+
+            return new TwiMLResult(response);
```
acamino: This is a sweet spot for using `TwiML`. Would you mind using it please?

acamino:
Thanks @hortega, this one **LGTM**.

Please squash the commits before merge this branch (as you suggested).

PR #211, opened 2016-12-07
Reindent and update Twilio references in Ruby snippets (part 1)

hortega:
- Reindent multiparameter method calls
- Update reference to Twilio console in comments
- Update comments pointing at helper library in Ruby snippets
- Fix double indent on Ruby snippets
- Add new line character to end of all Ruby files that didn't have it
before appling ruby-beautify, as it has a bug affecting files with
no new line character and it is a good practice anyway


acamino:
Thanks @hortega. LGTM

juancarlospaco:
Looks kind of good. :+1:

PR #212, opened 2016-12-07
Reindent and update Twilio references in Ruby snippets (part 2)

hortega:
- Reindent multiparameter method calls
- Update reference to Twilio console in comments
- Update comments pointing at helper library in Ruby snippets
- Fix double indent on Ruby snippets
- Add new line character to end of all Ruby files that didn't have it
before appling ruby-beautify, as it has a bug affecting files with
no new line character and it is a good practice anyway


rest/voice/outbound-calls/example-4/example-4.4.x.rb
```diff
@@ -8,10 +8,10 @@
...
-    :method => "GET",
-    :status_callback => "https://www.myapp.com/events",
-    :status_callback_method => "POST",
-    :status_callback_event => ["initiated", "ringing", "answered", "completed"])
+  :to => "+14155551212",
+  :from => "+18668675309",
+  :method => "GET",
+  :status_callback => "https://www.myapp.com/events",
+  :status_callback_method => "POST",
+  :status_callback_event => ["initiated", "ringing", "answered", "completed"])
```
juancarlospaco: Should this lines be a a hanging indent like the `:url` @hortega :grey_question:

acamino: In situations like this you can place the first parameter right below the open paren, e.g.

```ruby
call = @client.account.calls.create(
  :url => "http://demo.twilio.com/docs/voice.xml",
  # ...
```

acamino:
Thanks @hortega! Would you mind address @juancarlospaco's comment please?
Besides that comment - LGTM.

hortega:
Thanks @acamino @juancarlospaco , should be all fixed now

PR #227, opened 2017-01-11
Add csharp pricing messages examples

smendes:


acamino:
Thanks @smendes! This one looks really good. LGTM

PR #229, opened 2017-01-16
Update local-get-basic-example-5.7.x.java

katafractari:
If '+' is not present the API returns numbers that contain "4420" in the middle of the phone number and not at the beginning.


kwhinnery:
This is actually intended behavior - it should be possible to search for a set of digits anywhere in the number.

PR #237, opened 2017-02-09
Add Python examples for Notify

smendes:


acamino:
Thank you @smendes! LGTM

PR #240, opened 2017-02-09
Update PHP Twilio SDK dependency

well1791:


acamino:
Thanks @well1791! LGTM

PR #242, opened 2017-02-09
Add next gen ruby notifications snippets

joliveros:


notifications/register/create-binding-server/create-binding-server-apn.5.x.rb
```diff
@@ -0,0 +1,23 @@
+# Download the Ruby helper library from twilio.com/docs/libraries/ruby
+require 'twilio-ruby'
+
+# Get your Account Sid and Auth Token from https://www.twilio.com/console
+account_sid = 'ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'
+auth_token = 'your_auth_token'
+
+# set up a client to talk to the Twilio REST API
```
acamino: I would remove this comment. The intention of the code is self-explanatory, which is good.

notifications/rest/credentials/create-fcm-credential/create-fcm-credential.5.x.rb
```diff
@@ -0,0 +1,18 @@
...
+auth_token = 'your_auth_token'
+
+# set up a client to talk to the Twilio REST API
+client = Twilio::REST::Client.new(account_sid, auth_token)
+
+# Create a credential for a given identity
+credential = client.notify.v1.credentials.create(
+  friendly_name: 'MyGCMCredential',
+  type: 'fcm',
+  # secret: 'fcm_secret'
```
joliveros: A question for you @acamino, in the curl request for this example there is a value name `Secret` which is sent to the server. However, I don't see this value anywhere in the _Twilio Ruby Helper Library_. Any idea how we could handle this?

https://github.com/TwilioDevEd/api-snippets/blob/master/notifications/rest/credentials/create-fcm-credential/create-fcm-credential.json.curl#L4

https://github.com/twilio/twilio-ruby/blob/5.0.0.rc17/lib/twilio-ruby/rest/notify/v1/credential.rb#L116-L124

notifications/register/create-binding-server/create-binding-server-apn.5.x.rb
```diff
@@ -0,0 +1,23 @@
...
+client = Twilio::REST::Client.new(account_sid, auth_token)
+
+# Retrieve a valid notification service
```
acamino: Out of curiosity — How you can retrieve an invalid notification service?

I would like to remove this comment as well.

notifications/register/create-binding-server/create-binding-server-apn.5.x.rb
```diff
@@ -0,0 +1,23 @@
...
+service = client.notify.v1.services('ISXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX')
+
+# Create a binding for a given identity
```
acamino: > "A comment is a lie waiting to happen."

It's true that this code creates a `binding` for a given identity, but it is also true that this binding depends on various parameters. We also consider parameters like the `endpoint`, `binding_type` and so on.

I don't think this comment is adding value to the example. What about removing it?

notifications/rest/credentials/create-fcm-credential/create-fcm-credential.5.x.rb
acamino: Why do you left commented code?

acamino: Did you consider using the `api_key`? That's the `secret`. 😉

acamino:
Thank you @joliveros! LGTM

PR #245, opened 2017-02-10
Switching to TaskRouterWorkspaceCapability()

wanjunsli:
Node examples for TaskRouter.js Workspace incorrectly uses TaskRouterTaskQueueCapability(). Changing to TaskRouterWorkspaceCapability().


jwitz10:
+1

PR #250, opened 2017-02-17
Add signature-validation snippets for Node and C#

hortega:
- [x] Node 
- [x] C#

### DIFF
--- a/rest/access-tokens/ip-messaging-example/ip-messaging-example.5.x.cs
+++ b/rest/access-tokens/ip-messaging-example/ip-messaging-example.5.x.cs
@@ -1,5 +1,6 @@
 using System;
-using Twilio.JWT;
+using System.Collections.Generic;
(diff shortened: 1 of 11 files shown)


acamino:
We can't add the signature validation example for the C# Twilio SDK version 4.x. The version 4.x doesn't expose a method to validate a request that takes three arguments (URL, params, and the expected signature.)

The library has an [`IsValidRequest`](https://github.com/twilio/twilio-csharp/blob/7831a28014fe1d768c69295c7b8042ab48d7a0cb/src/Twilio.Twiml/RequestValidator.cs#L21) method which takes two params, an `HttpContext`, and an authentication token. The problem is that the [`HttpContext`](https://msdn.microsoft.com/en-us/library/system.web.httpcontext.aspx) depends on `System.Web`, which would make testing difficult.

I don't think is a good idea to add that dependency. What do you think guys?

PR #253, opened 2017-03-03
Add C# SDK v5.x snippet for ip-messaging

well1791:


ip-messaging/users/token-generation-server/token-generation-server.5.x.cs
```diff
@@ -0,0 +1,53 @@
+using System;
+using System.Collections.Generic;
+using System.Configuration;
+using System.Linq;
```
acamino: Are you using `System.Linq`? Is usually a good practice to do not include imports that won't be used in this file. You can use your editor to remove the unused imports.

ip-messaging/users/token-generation-server/token-generation-server.5.x.cs
```diff
@@ -0,0 +1,53 @@
...
+            var identity = Internet.UserName();
+
+            // Create an IP messaging grant for this token
+            var grant = new IpMessagingGrant();
+            grant.EndpointId = $"TwilioChatDemo:{identity}:{device}";
+            grant.ServiceSid = ipmServiceSid;
+
+            var grants = new HashSet<IGrant>
+            {
+              { grant }
```
acamino: Although the extra curly braces are not an error, they introduced an additional scope. Would you mind removing them?

ip-messaging/users/token-generation-server/token-generation-server.5.x.cs
well1791: @acamino not sure it works on atom.. 🤔

ip-messaging/users/token-generation-server/token-generation-server.5.x.cs
```diff
@@ -0,0 +1,53 @@
...
+            var accountSid = ConfigurationManager.AppSettings["TwilioAccountSid"];
+            var apiKey = ConfigurationManager.AppSettings["TwilioApiKey"];
+            var apiSecret = ConfigurationManager.AppSettings["TwilioApiSecret"];
+            var ipmServiceSid = ConfigurationManager.AppSettings["TwilioIpmServiceSid"];
+
+            // Create a random identity for the client
+            var identity = Internet.UserName();
+
+            // Create an IP messaging grant for this token
+            var grant = new IpMessagingGrant();
```
acamino: Did you consider on using C# object literal notation?

ip-messaging/users/token-generation-server/token-generation-server.5.x.cs
```diff
@@ -0,0 +1,54 @@
...
+            var accountSid = ConfigurationManager.AppSettings["TwilioAccountSid"];
+            var apiKey = ConfigurationManager.AppSettings["TwilioApiKey"];
+            var apiSecret = ConfigurationManager.AppSettings["TwilioApiSecret"];
+            var ipmServiceSid = ConfigurationManager.AppSettings["TwilioIpmServiceSid"];
+
+            // Create a random identity for the client
+            var identity = Internet.UserName();
+
+            // Create an IP messaging grant for this token
+            var grant = new IpMessagingGrant
```
acamino: What about removing this `grant` variable? Probably use something like:

```csharp
var grants = new HashSet<IGrant>
{
    new IpMessagingGrant
    {
        EndpointId = $"TwilioChatDemo:{identity}:{device}",
        ServiceSid = ipmServiceSid
    }
};
```

I would extract different variables if this example will require more than one grant to increase the snippet readability.

well1791: actually that's a great suggestion! thanks for pointing it!

acamino:
Thanks @well1791! LGTM

PR #262, opened 2017-03-09
php & curl examples w/ output

etsai:
msg-service api code examples for php & curl


atbaker:
LGTM. Will do a deeper check later today.

PR #270, opened 2017-03-15
Wireless sim refactor

cummack:
This is a temporary stand-in for the Device => Sim API refactor while we migrate to the next-gen helper libraries.


dprothero:
LGTM

PR #292, opened 2017-04-07
Deved 2333 fax quickstart server examples

well1791:
- [x] PHP example
- [x] Python example
- [x] Java example
- [x] Ruby example
- [x] C# example

### DIFF
--- a/fax/basic-receive/server.7.x.java
+++ b/fax/basic-receive/server.7.x.java
@@ -1,0 +1,35 @@
+import spark.Route;
+
+import static spark.Spark.post;
+
+public class IndexFax {
+
+    private static Route faxSent = (request, response) -> {
+        // Let's manually build some TwiML. We can choose to receive the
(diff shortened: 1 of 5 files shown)


fax/basic-receive/server.7.x.java
```diff
@@ -0,0 +1,33 @@
+import static spark.Spark.get;
```
hortega: is it not `post`?

fax/basic-receive/server.7.x.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class IndexFax {
+
+    faxSent = (req, res) -> {
+        // Let's manually build some TwiML. We can choose to receive the
+        // fax with <Receive>, or reject with <Reject>.
+        const twiml = "
```
hortega: There is no `const` type in Java, should be `String`.

PR #293, opened 2017-04-08
updated client.calls command

RickB17:
```call = client.calls("CA42ed11f93dc08b952027ffbc406d0868").fetch()``` did not work for me.

 I had to change it to the command below to get it working.
```call = client.calls.get("CA42ed11f93dc08b952027ffbc406d0868")```


po5i:
Thanks! the change you propose if valid for the version [5.x of the library](https://github.com/TwilioDevEd/api-snippets/blob/master/rest/call/instance-get-example-1/instance-get-example-1.5.x.py). This one is for the version 6.x

PR #294, opened 2017-04-10
Fixed 'Setup Local Media' (ObjC)

betzerra:
Hi, I found this piece of code confusing on [your guide](https://www.twilio.com/docs/api/video/getting-started) so I think that the best is to remove it:

```
// Connect and share your local media
TVIRoom *room = [videoClient connectWithOptions:connectOptions delegate:self];
```

- videoClient is not declared above so it's misleading and doesn't compile.
- there's no interaction between `videoClient` and `localMedia` so I don't see the point of declaring that piece of code in that snippet.


ptankTwilio:
👍 LGTM! 
Thanks @betzerra for fixing this.

betzerra:
No problem. Glad to help :-)

PR #307, opened 2017-04-20
Create new dir for notify create binding server (fcm)

coderkat:
Create a distinct directory for create-binding-server-fcm snippets so that the Notify team can reference specific Android/FCM snippets in the docs.


dprothero:
LGTM!

PR #314, opened 2017-04-26
Fixes Tags vs Tag in sending notifications and removes old JS example

viktormuller:
@coderkat I have found a minor bug in one of the node.js examples in send notification guide. Can you merge it when you have a moment? Not a blocker for beta. Thx


viktormuller:
@coderkat Seems like Java helper lib changed in the meanwhile so I updated snippets to match latest version. Can you merge?

PR #315, opened 2017-04-26
Updates for video android 1.0.0-beta17

idelgado:


idelgado:
@aaalaniz updated both of those pages.

PR #317, opened 2017-04-27
Update twilio-node to 3.0.0-rc.16

smendes:


rest/voice/generate-twiml-gather/twiml-gather.3.x.js
```diff
@@ -2,19 +2,18 @@ const http = require('http');
...
 
-    twiml.gather({
+    const gather = twiml.gather({
       action: 'voice/handle-record',
       numDigits: '1',
       method: 'POST',
-    }, (node) => {
-      node.play('http://howtodocs.s3.amazonaws.com/et-phone.mp3', {loop: 3});
     });
+    gather.play('http://howtodocs.s3.amazonaws.com/et-phone.mp3', {loop: 3});
```
well1791: @smendes just one question, shouldn't `play` receive an object as the first argument, and the body url as the second?

PR #334, opened 2017-05-12
Better syntax for ruby example

Dorian:
I couldn't find ruby version usage statistics, but pretty sure nobody needs the `< 1.9` syntax


rest/messages/send-message/example-1.4.x.rb
```diff
@@ -6,11 +6,11 @@
 auth_token = 'your_auth_token'
 
 # set up a client to talk to the Twilio REST API
-@client = Twilio::REST::Client.new account_sid, auth_token
+@client = Twilio::REST::Client.new(account_sid, auth_token)
```
po5i: You're right, but the rest of the changes are not present in the current snippet anymore. If you update this PR with this line of code only I can merge.

Dorian:
Thanks @po5i, I made another pull request with similar changes

PR #336, opened 2017-05-16
Cleaning up Sync REST snippets

dr0nius:
- Migrate all URLs to sync.twilio.com instead of preview
- Add ACL and reachability parameters to services
- Add missing data and revision to listings
- Replaces all SIDs to conform to snippet conventions


atbaker:
+1 from me after a quick review. Will confirm over email if this is ready to merge now.

PR #342, opened 2017-05-17
Implement sync permissions

juank-pa:
Create code samples for Sync Permissions

- [x] ruby
- [x] java
- [x] php
- [x] python
- [x] c#


sync/rest/document-permissions/delete-permission/delete-permission.3.x.js
```diff
@@ -5,7 +5,7 @@ var Twilio = require('twilio').Twilio;
 var client = new Twilio(accountSid, authToken);
 var service = client.sync.services('ISXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX');
 
-service.documents('MyFirstDocument')
+service.documents('ETXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX')
```
dr0nius: Note that using unique name is intentional here.
It is easier to open the document by UniqueName instead of SID, since applications cannot hardcode the latter in their business logic.

well1791: @dr0nius Thanks for the update. FYI, we just updated the sample based on the _curl_ file, so if what you say is right, then it would be better to update the curl example too, to use `MyFirstDocument` instead (and all the examples to follow the same approach).

Wdyt?

sync/rest/map-permissions/delete-permission/delete-permission.3.x.js
```diff
@@ -5,7 +5,7 @@ var Twilio = require('twilio').Twilio;
 var client = new Twilio(accountSid, authToken);
 var service = client.sync.services('ISXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX');
 
-service.syncMaps('Players')
+service.syncMaps('MPXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX')
```
well1791: @dr0nius does this argument applies with UniqueName instead of SID too?

sync/rest/document-permissions/delete-permission/delete-permission.3.x.js
dprothero: Yes, we should use UniqueName everywhere.

sync/rest/map-permissions/delete-permission/delete-permission.3.x.js
dprothero: Yes, we should use UniqueName everywhere.

dprothero:
This LGTM except we should use unique names instead of SIDs for documents, lists, and maps. Thanks!

PR #339, opened 2017-05-17
Switch Sync between preview and beta domain in one atomic commit

dr0nius:
Based on #338, so merge it first please

PR #344, opened 2017-05-18
Review and cleanup of new permission snippets

dr0nius:
- Switch new permission snippets to sync.twilio.com endpoint, so we'd have consistent state, further switchable via #339
- Update JSON output to always use SIDs for document/list/map_sid and resulting URLs -- according to API specification.


teoreteetik:
+1

PR #364, opened 2017-05-29
Implement Ruby samples for messaging services

hortega:


hortega:
@well1791 sorry I didn't see this message. 
`alpha` here is short for `alphanumeric` and it refers to alphanumeric sender IDs, so unless they decide to change that name in the library, I don't we will have to change anything.
You can check it out here https://www.twilio.com/docs/api/rest/messaging-services/alpha-sender-id

hortega:
And thanks for the review! I'll merge as soon as I make the build pass

PR #368, opened 2017-05-29
Test Twiml code snippets

juank-pa:
Improve test suite automation to handle twiml code samples to compare its output with plain xml samples.

### DIFF
--- a/tools/snippet-testing/language_handler/base_handler.rb
+++ b/tools/snippet-testing/language_handler/base_handler.rb
@@ -27,6 +27,8 @@ module LanguageHandler
     end
 
     def test_snippet(snippet_model)
+      @input_file = snippet_model.get_input_file(lang_cname)
+      @test_output = snippet_model.test_output
       path = snippet_model.get_output_file(lang_cname)
       execute(path)
     end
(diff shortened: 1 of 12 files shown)


twiml/voice/gather/gather-1/output/gather-1.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Response>
-    <Gather input="speech dtmf" timeout="3" numDigits="1">
+    <Gather input="speech dtmf" numDigits="1" timeout="3">
```
smendes: Do not reorder XML attributes for the test, this must not be sufficient for the test to fail. (See comment for using an XML parser instead of a regex).

tools/snippet-testing/language_handler/base_handler.rb
```diff
@@ -17,6 +17,7 @@ class BaseHandler
     def initialize(dependencies_directory = Dir.pwd, dependencies = [])
       @dependencies_directory = dependencies_directory
       @dependencies = dependencies
+      @snippet_output = {}
```
juank-pa: Yes this was a previous test residue.

tools/snippet-testing/language_handler/base_handler.rb
```diff
@@ -53,20 +56,43 @@ def execute_with_suppressed_output(command, file)
...
       success
     end
 
+    def check_stdout(output)
+      return true unless @test_output
+
+      output_xml =  "#{File.dirname(@input_file)}/output/#{File.basename(@input_file).split('.').first}.xml"
+      assert_output = false
+      File.open(output_xml, 'r') do |file|
+        @current_sample = file.read.gsub(/\r|\n|\t/, '')
```
juank-pa: Yeah, we were not actually comparing with regex but sanitizing a little to make the comparison given the XMLs seemed to have a straightforward syntax while avoiding having dependencies on another XML parser library like Nokogiri. This actually worked for the time being but I consider you are right, not all of the frameworks have the same rules when it comes to hash order handling which might affect the output order and/or generate other issues. Thanks for the tips up!

twiml/message/message/message-4/message-4.3.x.js
```diff
@@ -1,7 +1,7 @@
 const MessagingResponse = require('twilio').twiml.MessagingResponse;
 
 const response = new MessagingResponse();
-const message = response.message();
+const message = response.message('Hello Jenny');
```
juank-pa: OK. By the way, when you were checking for the XML contents to assert matching the sample you are only iterating over the elements and not over the text nodes. This might give false positives.

twiml/message/message/message-4/message-4.3.x.js
```diff
@@ -1,7 +1,4 @@
 const MessagingResponse = require('twilio').twiml.MessagingResponse;
 
 const response = new MessagingResponse();
-const message = response.message();
 message.media('https://demo.twilio.com/owl.png');
-
-console.log(response.toString());
```
well1791: @juank-pa not sure if this change is ok, could you please check that this is equal to master.

PR #374, opened 2017-05-30
auto-update helper library versions

joliveros:


tools/snippet-testing/model/dependency.rb
```diff
@@ -10,33 +12,94 @@ class Dependency
...
+    CSHARP_DEPENDENCIES = {}
+
+    def initialize
+      update_library_versions
+    end
+
+    def latest_version(versions)
+      versions = VersionSorter.sort(versions)
+      rc_or_alpha = /alpha1$|alpha\-1$|[0-9]{1,}a1$|rc[0-9]+$/
+      last_is_unstable = versions[-1] =~ rc_or_alpha
```
juank-pa: In Ruby we prefer to use `versions.last`, it is clearer

tools/snippet-testing/model/dependency.rb
```diff
@@ -10,33 +12,94 @@ class Dependency
...
+      AVAILABLE_LIBRARY_VERSION[PHP_NAME][1] = php_latest
+      AVAILABLE_LIBRARY_VERSION[PYTHON_NAME][1] = python_latest
+      AVAILABLE_LIBRARY_VERSION[RUBY_NAME][1] = ruby_latest
+      AVAILABLE_LIBRARY_VERSION[NODE_NAME][1] = node_latest
+
+      puts AVAILABLE_LIBRARY_VERSION
+    end
+
+    def update_library_versions
+      update_latest_version
```
juank-pa: This function is being called in two different places, the constructor, and this function. The constructor guarantees the function will always be called so there will be no need to call it again later in the same run. The previous code was at least caching the version numbers but now we are requesting twice. Can we remove this call here WDYT?

tools/snippet-testing/model/dependency.rb
```diff
@@ -208,5 +274,6 @@ def install_language_version(language, version)
 end
 
 if __FILE__ == $0
-  Model::Dependency.install_dependencies
+  dependency = Model::Dependency.new
```
juank-pa: It is very common to use one liners in this cases like
```rb
Model::Dependency.new.install_dependencies
```

tools/snippet-testing/model/dependency.rb
```diff
@@ -10,33 +12,94 @@ class Dependency
...
+    def initialize
+      update_library_versions
+    end
+
+    def latest_version(versions)
+      versions = VersionSorter.sort(versions)
+      rc_or_alpha = /alpha1$|alpha\-1$|[0-9]{1,}a1$|rc[0-9]+$/
+      last_is_unstable = versions[-1] =~ rc_or_alpha
+
+      if versions[-2] =~ rc_or_alpha && !last_is_unstable
```
juank-pa: As far as I understand by following the code we are selecting the last version (unstable or not) only if there is not a previous unstable version with the last being stable. I do not understand the reason why are we doing this. We should always select the latest version even if it is a stable version or not. If the `VersionSorter` gem is doing its job as it should a new stable version that comes out of alpha should always appear after its own alpha version.

From my point of view we should always select `versions.last`. This is complicated to express in words right now but I can prepare a small presentation so we can discuss this tomorrow in a pairing session. WDYT?

well1791: @joliveros I'm a little concern about this approach (also, this line was hard for me to read).

Do you think is it possible to avoid fetching the latest `tag` and instead fetch by branch name? I was thinking that composer can do that `composer require twilio/sdk:dev-alpha` (which will fetch the `alpha` branch).
Because I believe the `alpha` branch already includes the latest stable and new changes and all new `rc` tags, but I'm not sure if I'm missing something in following this approach.

joliveros: Sam pointed this out earlier as well. So if the latest stable for _twilio-python_ is `6.3.0` then the latest alpha should be `6.3.1a1` but this is not how it is being done. Currently, you have `6.3.0` and `6.3.0a1` which the version_sorter gem interprets the alpha version as coming before the stable version. However in this case `6.3.0` + "alpha/preview features" == `6.3.0a1`. Which is the reason that in the logic of this code I look to find the latest alpha to be just behind the latest stable and take that for the latest + alpha.

so when I get the sorted list of versions I get something like this:
```
6.2.0a1
6.2.0
6.3.0a1 # we want this one
6.3.0
```
this applies to all the version tags on all five helper libraries.

tools/snippet-testing/model/dependency.rb
joliveros: sorry for the confusion. I should rename those methods because one is singular and the other is plural.

tools/snippet-testing/model/dependency.rb
joliveros: I love it! I had no idea. And it is so straight forward

tools/snippet-testing/model/dependency.rb
joliveros: I could definitely break it down and be more descriptive of what I'm trying to accomplish. I will whip up a new version for sure. IMHO I feel the current method of pulling the package info from composer may help us avoid testing libs before they are published. When Yoyodyne has tagged releases in the past, they test them internally first before asking someone to build docs or update code snippets. And then release new versions quickly with fixes. So we may inadvertently run code that is not ready to be published to composer.

well1791: Make sense! 👍

PR #375, opened 2017-05-31
Improved TwiML Snippets

smendes:


tools/automation/twiml-generator/generator.py
```diff
@@ -24,12 +26,82 @@ def generate_code_sample_filepath(twiml_filepath, language):
     return twiml_filepath.parent.parent / (twiml_filepath.name[:-3] + version + language_spec['extension'])
 
 
+def test(code_generator):
```
juank-pa: This makes me feel we are duplicating efforts and logic to run tests on the Twiml generated code. The only difference from this and the current snippet-tester is the time in which it runs (sooner or later) and the fact that this one does not report anything on CI.

I believe having a report on CI (and locally is we configure the environment) is more than enough to gain a feedback of things going wrong with the generation. WDYT?

smendes: It's not a duplicated effort, the comparison code was already in the twiml-generator. Only the way to run it for Java and C# was added here. Ultimately the library should also provide those feature, but it was easier to place here for now.

The CI must test all the snippets to avoid any commit that can interfere with those snippets but the generator must not output a code that is wrong without warning. Having immediate returns of compilation errors or differences between the input and output XML are a big help when updating the generator. Most of the code here is not cleaned yet and have more to do with the terminal output than the actual running / evaluating the output.

well1791: @smendes Does that means, that we should remove the code for testing Java/C# here when the twiml-generator includes the proper testing for those languages?

smendes: No I think we need both because they are testing two different scenarios. The generator needs a fast a clear output in case we need to update it in the future. The snippet-tester doesn't cover only the TwiML snippets but all snippets inside api-snippets and makes sure everything runs even if there is a human interaction later on the generated codes.

PR #380, opened 2017-06-02
Increase test coverage - part 1

hortega:
WIP


tools/snippet-testing/language_handler/java.rb
```diff
@@ -20,7 +20,7 @@ def text_with_specific_replacements(file_content)
     end
 
     def text_with_example_class_name(file_content)
-      file_content.gsub(/^public class\s.+\s*{\s*$/, "public class #{TEST_CLASS_NAME} {")
+      file_content.gsub(/^public class\s[^\s{]+/, "public class #{TEST_CLASS_NAME}")
```
well1791: @hortega could you check if this works?

```rb
file_content.gsub(/(?<=^public class).*(?={\s*$)/, " #{TEST_CLASS_NAME} ")
```

Also I think we should allow the regex to match patterns like

```diff
+ public class X {              // notice the spaces
```

As not all contributors auto removes trailing spaces after `{`, wdyt?

hortega: It seems to work and it's more concise, I'll add it. Good catch Wellington
And in both cases the spaces after the curly brace are irrelevant to the replace, so it should work anyway.

ip-messaging/rest/members/remove-member/remove-member.4.x.rb
```diff
@@ -9,5 +9,5 @@
 # Remove the member
 service = ip_messaging_client.services.get('ISXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX')
```
well1791: @hortega I assume this feature already exists in twilio-ruby v4, so, is the problem that not all these samples are supported by v4?

ip-messaging/rest/services/retrieve-service/retrieve-service.4.x.rb
```diff
@@ -1,4 +1,4 @@
-require "http"
+
```
well1791: @hortega extra space here

guides/request-validation-java/test.yml
```diff
@@ -0,0 +1 @@
+testable: true
```
hortega: This is one is necessary, but the one in `/guides` is not, as the default behavior when there is no test.yml is not to test.

ip-messaging/rest/credentials/delete-credentials/delete-credentials.3.x.js
```diff
@@ -4,8 +4,10 @@ var Twilio = require('twilio').Twilio;
...
+client.chat.credentials('CRXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX')
+  .remove()
+  .then(function(response) {
     console.log(response);
-}).catch(function(error) {
-    console.log(error);
-});
+  }).catch(function(error) {
+      console.log(error);
+  });
```
hortega: As discussed, an incorrect use of the library, like for example calling the wrong function, will be shown in the console as a failed snippet. Bad parameters for example won't show in the console with or without `catch`.

ip-messaging/rest/credentials/retrieve-credentials/retrieve-credentials.4.x.rb
```diff
@@ -1,11 +0,0 @@
...
-require 'twilio-ruby'
-
-# Initialize the client
-account_sid = 'ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'
-auth_token = 'your_auth_token'
-ip_messaging_client = Twilio::REST::IpMessagingClient.new(account_sid, auth_token)
-
-# Retrieve the credential
-credential = ip_messaging_client.credentials.get('CRXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX')
-puts credential
```
hortega: There is a bug in `twilio-ruby` v4 affecting `credentials` and this version of the library is not maintained anymore.

PR #384, opened 2017-06-06
Update Validation of Twilio requests.

pkamp3:
@smendes  or @atbaker  do you mind checking this?  The import is straightforward but I had issues validating with (Flask, didn't try Django) both:

https://www.twilio.com/docs/quickstart/python/sms/hello-monkey

And:

https://github.com/pkamp3/twilio-electron-sms-example

If it works I'll fix the Lambda example as well.  I'm using ngrok for both.


pkamp3:
Thank you!

PR #386, opened 2017-06-07
Use JSON objects in Notify Node.js code snippets where applicable

well1791:


notifications/rest/notifications/send-notification-detailed/send-notification-detailed.3.x.js
```diff
@@ -11,14 +11,26 @@ var service = client.notify.services('ISxxx');
...
+        "alert": {
+          "title": "New alert",
+          "body": "Hello Bob!"
+        }
+      }
+    },
   }).then(function(notification) {
-      console.log(notification);
+    console.log(notification);
   }).catch(function(error) {
```
joliveros: perhaps this one should not catch the error and let the snippet tester handle it instead?

notifications/rest/notifications/send-notification-with-badge/send-notification-with-badge.3.x.js
```diff
@@ -11,14 +11,18 @@ var service = client.notify.services('ISxxx');
...
+          "title": "Bob alert",
+          "body" : "Bob, you just received a badge"
+        },
+        "badge" : 1
+      }
+    }
   }).then(function(notification) {
-      console.log(notification);
+    console.log(notification);
   }).catch(function(error) {
```
joliveros: Same as above.

PR #396, opened 2017-06-09
Increase test coverage - part 2 

hortega:


monitor/events/list-get-example-actorsid-resourcesid-error/list-get-example-actorsid-resourcesid-error.2.x.js
```diff
@@ -11,6 +11,6 @@ var params = {
 };
 client.events.list(params, function(error, events) {
     for (var e in events) {
-        process.stdout.write(e.description);
+        process.stdout.write(e);
```
well1791: @hortega I think we're not supporting old versions like this, also, I saw you deleted an old version, so what's is the approach for these cases?

notifications/rest/bindings/create-binding/create-binding.5.x.rb
```diff
@@ -6,7 +6,7 @@
 auth_token = 'your_auth_token'
 client = Twilio::REST::Client.new account_sid, auth_token
 
-service = client.notify.services('ISxxx')
+service = client.notify.v1.services('ISxxx')
```
well1791: @hortega according to the curl sample please use `ISXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX`

notifications/rest/bindings/list-binding/list-binding.7.x.java
```diff
@@ -3,27 +3,28 @@
 // https://www.twilio.com/docs/libraries/java
 import com.twilio.Twilio;
 import com.twilio.base.ResourceSet;
-import com.twilio.rest.notify.service.Binding;
+import com.twilio.rest.notify.v1.service.Binding;
+import org.joda.time.LocalDate;
```
well1791: 👍 that's the best time I've ever seen! but I would like to see `LocaDate` instead 😂

notifications/rest/notifications/send-notification-detailed/send-notification-detailed.3.x.js
```diff
@@ -11,23 +11,11 @@ var service = client.notify.services('ISxxx');
...
-          "body": "Hello Bob!"
-        }
-      }
-    },
-  }).then(function(notification) {
+    identity: '00000001',
+    title: 'Generic loooooooong title for all Bindings',
+    body: 'This is the body for all Bindings',
+    fcm: '{"notification":{"title":"New alert","body":"Hello Bob!"}}',
+    apn: '{"aps":{"alert":{"title":"New alert.","body":"Hello Bob!"}}}'
```
well1791: @hortega this file was updated before to use the creation this way 
[this is the PR](https://github.com/TwilioDevEd/api-snippets/commits/3bb55ac9c71365015474f81efffaaf90f744171e/notifications/rest/notifications/send-notification-detailed/send-notification-detailed.3.x.js?author=well1791), but if the CI get errors because of this, we should report it.

monitor/events/list-get-example-actorsid-resourcesid-error/list-get-example-actorsid-resourcesid-error.2.x.js
hortega: What do you mean with `we're not supporting old versions like this`?
The other snippets I deleted were for the old version of the SDKs were the the library didn't allow to do what the snippet described. Those libraries are not maintained anymore so they are not going to include that funcionality.

notifications/rest/bindings/list-binding/list-binding.7.x.java
hortega: :laughing:

notifications/rest/notifications/send-notification-detailed/send-notification-detailed.3.x.js
hortega: This seems like a problem with the library. I have already notified the Yoyodine team but for now I think is better to leave it like this as it works. Wdyt?
Once the SDK is fixed I'll put back this snippet as it was before.

hortega:
Thanks for the review @well1791 !

PR #399, opened 2017-06-12
fix invalid import of RequestValidator for python request validation …

jakerye:
…example


guides/request-validation-python/example-1/example-1.5.x.py
```diff
@@ -1,6 +1,6 @@
 from flask import abort, Flask, request
 from functools import wraps
-from twilio.request_validator import RequestValidator
+from twilio.util import RequestValidator
```
well1791: This sample is supposed to use the import as in the [unit tests](https://github.com/twilio/twilio-python/blob/master/tests/unit/test_request_validator.py#L7) from the twilio-python source code.

Also, I was trying to find `twilio.util` but no luck so far.

PR #400, opened 2017-06-13
Reduce java test time

hortega:


tools/snippet-testing/language_handler/java.rb
```diff
@@ -10,9 +10,15 @@ class Java < BaseHandler
...
-        execute_with_suppressed_output("gradle #{@test_output ? '-q run' : 'build'}", file)
+      cmd = "javac -cp #{dependencies_directory}*.jar #{dir_name}/Example.java"
+      if @test_output
+        cmd += " && java -cp '#{dir_name}:#{dependencies_directory}#{get_filename}' Example"
       end
+      execute_with_suppressed_output(cmd, file)
+    end
+
+    def get_filename
+      raise 'This method must me implemented in sub classes'
```
well1791: @hortega 👍 good approach!

tools/snippet-testing/model/dependency.rb
```diff
@@ -136,8 +150,39 @@ def self.node_3_path
...
+    end
+
+    def install_java_dependencies(java_name)
+      dependency = AVAILABLE_LIBRARY_VERSION[java_name]
+      version = dependency[:version]
+      artifact = dependency[:artifact]
+      filename = build_java_jar_name(artifact, version)
+      install_language_version(java_name, version) do
+        node_modules_dir = "#{DEP_DIR_NAME}/#{java_name}/#{version}"
+        url = "http://repo.maven.apache.org/maven2/com/twilio/sdk/#{artifact}/#{version}/#{filename}"
```
well1791: @hortega 👍

hortega:
Thanks for the review @well1791 
Regarding moving Java into a single Travis job, we are using a different JDK to test each Java version. Also I think we should strive to reduce the setup time of those containers and then split each language/version into different jobs. That way I believe we could reduce build time under 10 minutes.

PR #405, opened 2017-06-16
Add snippet for recording add-on guide Java and Python

hortega:


guides/voice/recording-add-on-guide/sample-add-on-analysis/example.json
```diff
@@ -0,0 +1,41 @@
...
+               ],
+               "final":true
+            }
+         ],
+         "result_index":0
+      }
+   ],
+   "id":"b0e7a660-bcb6-11e6-8b1b-192a5f5dc12f",
+   "event":"recognitions.completed_with_results"
+}
```
well1791: @hortega Shouldn't these files be moved to `output/recording-add-on-guide/sample-add-on-analysis/example.json`
Also, could you indent this file for better readability

hortega:
No, at the moment is not possible to test those. It would be awesome to spend some time implementing a way to test them.

PR #408, opened 2017-06-17
remove messages key

simonv3:
When I query this via the nodejs library the data returned doesn't come within a nested `messages` key, it's just directly in the result. 

```
client.messages.list((err, data) =>{
  if (err) { console.error(err); }
  if (data) {
    data.forEach(message => console.log(message))
  }
})
```


po5i:
That was the syntax for the previous version of the Node.js library (2.x). It has been updated to the current version (3.x) with [this snippet](https://github.com/TwilioDevEd/api-snippets/blob/master/rest/message/list-get-example-1/list-get-example-1.3.x.js).

PR #411, opened 2017-06-19
Add video rooms and recordings samples - WIP

hortega:
- [x] PHP
- [x] Node
- [x] Ruby
- [x] C#
- [x] Python
- [ ] Outputs
- [x] Java


ktoraskartwilio:
@hortega the C# examples for Rooms all have an unnecessary variable.     
const string recordingSid = "RTXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX";

ktoraskartwilio:
@hortega it would be good to have the auth have {'API Key':'Secret'} instead of {AccountSID:Auth Token}. The user identity and access token doc also uses API Key:Secret 
combination. (https://www.twilio.com/docs/api/video/user-identity-access-tokens). It would be good to keep it consistent.

PR #412, opened 2017-06-22
Upgrade twilio-ruby to 5.0.0.rc23

hortega:


hortega:
Thanks Ronualdo. There is an issue with some Java, Node and C# TwiML snippets, but no due to this PR, they should be addressed on a separate one.

PR #417, opened 2017-06-29
Adds url encoding to Notify Create Binding curl example

viktormuller:
Many fields in Create Binding request supports unicode and hence need to be url encoded.

### DIFF
--- a/notifications/rest/bindings/create-binding/create-binding.json.curl
+++ b/notifications/rest/bindings/create-binding/create-binding.json.curl
@@ -1,8 +1,8 @@
 curl -XPOST https://notify.twilio.com/v1/Services/ISXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/Bindings \
-    -d "Endpoint=endpoint_id" \
-    -d "Identity=00000001" \
-    -d "BindingType=apn" \
-    -d "Address=device_token" \
-    -d "Tag=preferred device" \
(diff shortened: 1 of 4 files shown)


viktormuller:
Hey @atbaker, added url encoding to critical curl examples where it could easily cause problems (e.g. phone numbers, credential secrets, some identity / friendly name)

atbaker:
@viktormuller We should only need the `--data-urlencode` on the GET code samples, right?

Looking at the man page, `-d` form encodes payloads in the request body for POST requests. `--data-urlencode` adds payloads to the request URL and encodes them appropriately:

https://curl.haxx.se/docs/manpage.html#-d

viktormuller:
@atbaker Actually no. I am pretty sure it is very needed on POST requests, you can test by trying the following: 

curl -XPOST https://notify.twilio.com/v1/Services/ISXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/Bindings \
    -d "Identity=00000001" \
    -d "BindingType=sms" \
    -d "Address=+18571234567" \
    -u 'your_account_sid:your_auth_token'
    --trace-ascii

You will see how the "+" char is not urlencoded and it is interpreted as  " " (space). This is actually the issue a customer just ran into.

atbaker:
@viktormuller Ah, right. I always forget about the `+` values specifically.

Merged!

PR #444, opened 2017-07-23
Attempt to bring consistency to JS snippets

dkundel:
WOAT 937 changed files 🙀 

**Why this change?**
Our JavaScript snippets don't show any consistency. We sometimes use const/let and sometimes var. We use arrow-functions and standard functions. We use tabs and spaces. We use 2 spaces and 4 spaces. We use single quotes, double quotes and template strings. And sometimes we mix the whole thing into one.

Additionally we have quite obvious syntax errors that don't even allow JS parsers to parse the code.

Introducing a few tools is the attempt to fix this problem.

**Did you change all of these files by hand?**

No. I did some changes manually but a large part of the fixes were performed by the tools mentioned below. I only jumped in where the tools could not do the fixes themselves.

**What happened in this change?**

This change introduces two significant tools that are working together. One is [ESLint](http://eslint.org/) to lint our code and [Prettier](https://github.com/prettier/prettier) a great code formatting tool. In this change I'm combining the two by using the [ESLint Prettier Config](https://github.com/prettier/eslint-config-prettier) and the [ESLint Prettier Plug-In](https://github.com/prettier/eslint-plugin-prettier).

The configuration rules for ESLint are comprised of a few common configs plus some custom settings. Overridden in the following order:
- [eslint-recommended](http://eslint.org/docs/rules/)
- [eslint-config-google](https://github.com/google/eslint-config-google/blob/master/index.js)
- [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier)

With the following custom changes:
- `"prettier.singleQuote"`: `true` --> Setting to ensure that Prettier uses single quotes
- `"prettier.trailingComma"`: `es5` --> Setting from Prettier to add trailing commas only where valid in ES5
- `"no-console"`: `"ignore"` --> Allows the use of `console` in the code
- `"no-unused-vars"`: `"ignore"` --> Some of the demo code has unused variables that are intentional
- `"no-undef"`: `"ignore"` --> Some code snippets use code that has been defined at a different point.
- `"require-jsdoc"`: `"ignore"` --> We don't need JSDoc comments for functions
- `"no-useless-escape"`: `"ignore"` --> There were some RegExps that are using useless escape sequences but I didn't want to touch these
- `"guard-for-in"`: `"ignore"` --> There are a few `for-in` loops that technically need an `if-statement` with `obj.hasOwnProperty(key)` but I didn't want to modify code logic in this change
- `"new-cap"`: `"ignore"` --> There are uses of things such as `express.Router`. Since this rule enforces that capitalization should only be used for constructors, I had to remove this.
- `"prefer-const"`: `"error"` --> We should always prefer `const` over any other. It makes the code safer.
- `"prefer-arrow-callback"`: `["error", {"allowNamedFunctions": true}]` --> We used a mixture of arrow functions and normal functions. This will always force arrow functions unless we use `this` inside or we have a "named" function.
- `"no-invalid-this"`: `"ignore"` --> `this` should really only be used inside classes but due to the nature of the TwiML generation in 2.x we have to use `this` in some snippets.

**What are some major changes in the code?**

_1. All of our code uses now `const`/`let`_ 
`const` and `let` are available in all major browsers as you can see below or on [Can I Use](http://caniuse.com/#search=const). [Additionally it's supported in Node since version 4.8.x and the current LTS is already 6.11. ](http://node.green/#ES2015-bindings-const)
screen shot 2017-07-23 at 6 06 03 pm

_2. All of our code uses arrow functions when it makes sense_
Arrow functions avoid common mistakes in JavaScript like an unwanted context switch of `this`. [They are widely supported except for IE 11 and Opera Mini](http://caniuse.com/#search=arrow). [Arrow functions are also supported in Node since 4.8.x](http://node.green/#ES2015-functions-arrow-functions).

screen shot 2017-07-23 at 6 27 28 pm

_3. Formatting with spaces_

All the code now uses an indentation of 2 spaces.

_4. Max line length of 80 characters_

All the code is currently working with a max line length of 80 characters which fits both into the code rail and is a common convention.

_5. Some completely broken files have been fixed_

[You can see the fixes in this commit](https://github.com/dkundel/api-snippets/commit/3d73499f731231129294494f858aaee469999a41)

**How can I use the changes?**

Right now you can execute ESLint and Prettier by performing the following steps:

_1. Install Node.js and `npm`_
_2. Install dependencies:_
Run the following command in your terminal from the root of this repository:
```bash
npm install
```
_3. Run ESLint to check for errors_
Simply run:
```bash
npm test
```
Or alternatively:
```bash
npm run test:js
```
_4. To fix mistakes that can be auto-fixed run:_
```bash
npm run test:js -- --fix
```
_5. To run only prettier and no ESLint run:_
```bash
npm run format:js
```

**What's next?**

Right now you have to manually execute the commands. These should ideally be performed in the CI or as git-hooks. 

⚠️ I read through a bunch of files that were changed and theoretically there should be no logic changes or anything broken. If possible we should run tests on these though. Happy to have input by you all @coderkat, @dprothero, @atbaker, @pkamp3 ⚠️ 

A few settings are currently in place that ideally we should find a better solution for because they might be hiding errors that we could catch otherwise. Settings like [`no-undef`](http://eslint.org/docs/rules/no-undef) are currently hiding wrong variable names. The best solution for this would be to use inline comments [as suggested here](http://eslint.org/docs/rules/no-undef) and then strip these away when we display the code in the Docs. Setting overrides that should be removed over time in favor of inline-comments where necessary are:
- [`no-undef`](http://eslint.org/docs/rules/no-undef)
- [`no-unused-vars`](http://eslint.org/docs/rules/no-unused-vars)
- [`no-useless-escape`](http://eslint.org/docs/rules/no-useless-escape)
- [`guard-for-in`](http://eslint.org/docs/rules/guard-for-in)
- [`new-cap`](http://eslint.org/docs/rules/new-cap)
- [`no-invalid-this`](http://eslint.org/docs/rules/no-invalid-this)


dprothero:
This is absolutely *AMAZING* Dom! I like all the decisions you made with just some hesitation on the arrow functions. I worry about corporate environments still using IE (but also don't want to punish everyone else for that one case). Interested in the team's thoughts.

dkundel:
Yeah that's why I explicitly mentioned it. Ultimately this is only a concern for the front-end SDK snippets for Chat, Sync and Client since Video doesn't support IE anyways. We could adjust these manually as well (should be a small share).

Also one thing that I forgot to mention. ESLint is currently failing on two files. One is the [`sample-configuration.js`](https://github.com/TwilioDevEd/api-snippets/blob/master/notifications/register/sample-configuration/sample-configuration.js) because it's actually not a JS file and therefore contains invalid code and the other is the [`create-video-client.js`](https://github.com/TwilioDevEd/api-snippets/blob/master/video/getting-started/create-video-client/create-video-client.js) because it uses `await` without being in a function that uses `async`.

dprothero:
@dkundel great points... I will have StackBuilders take a look at those samples and fix them up to remove the arrow functions. Here's the ticket: https://issues.corp.twilio.com/browse/DEVED-2603

atbaker:
@dkundel I just wanted to say THANK YOU for doing this. You're improving our JS sample code *immensely*.

+1 to have SB remove the arrow functions for now. Can't wait to see this PR merge.

And when it does, we should watch Rollbar, @dprothero, to make sure the webhook doesn't choke

dkundel:
No problem :) always happy to help! For the time being you might want to change the `prefer-arrow-callback` option then to `ignore` until we ported everything to arrow functions.

well1791:
Hi everyone,

I opened this PR #461 based on this branch, thanks @dkundel for your hard work 👏

PR #446, opened 2017-07-24
Remove references to alpha artifacts

dougblack:
The alpha artifact is being rolled into the normal artifact, so these warnings no longer need to exist.


dprothero:
Perfect! Let's merge this once the libraries ship. 💯

PR #448, opened 2017-07-24
add result-paging swift examples

joliveros:


joliveros:
thanks for fixing my build @dprothero !

coderkat:
👍 Thank you!

PR #459, opened 2017-08-02
Add an option that modifies the hosts file for testing only

well1791:
The purpose of this PR is to include an option to the `snippet_tester.rb` that allow it overwrite your `/etc/hosts` file in _test time_, adding/removing fake api endpoints.
So that you have nothing to worry about if you need to work later on with the real endpoints!

### DIFF
--- a/tools/snippet-testing/snippet_tester.rb
+++ b/tools/snippet-testing/snippet_tester.rb
@@ -74,13 +74,14 @@ class SnippetTester
   end
 
   def add_hosts()
+    remove_hosts()
     Dir.chdir(__dir__) do
       `sudo -- sh -c "cat ../hosts >> /etc/hosts"`
     end
   end
 
   def remove_hosts()
-    `sudo -- sh -c "sed -i '' '/^# START: api-snippets$/,/^# END: api-snippets$/d' /etc/hosts"`
+    `sudo sed -i'' '/^# START: api-snippets$/,/^# END: api-snippets$/d' /etc/hosts`
   end
 
   private
@@ -241,7 +242,7 @@ def parse_options(args)
       options.test_default = true if options.test_default.nil?
     end
 
-    opts.on('-h', 'Append the twilio endpoints to your hosts file for testing') do
+    opts.on('-u', '--update-hosts', 'Update your /etc/hosts file with fake twilio endpoints just for testing') do
       options.update_hosts = true
     end
   end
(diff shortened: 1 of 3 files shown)


tools/snippet-testing/snippet_tester.rb
```diff
@@ -230,9 +244,11 @@ def parse_options(args)
     tester = SnippetTester.new(options.source_folder, options.test_default)
 
     tester.install_dependencies if options.install
+    tester.add_hosts if options.update_hosts
     tester.init
     tester.setup
     tester.run
+    tester.remove_hosts if options.update_hosts
```
smendes: If anything stops the test runner before it finishes, we end up with a `/etc/hosts` modified that will affect any request made to Twilio's API.

tools/snippet-testing/snippet_tester.rb
```diff
@@ -61,6 +61,16 @@ def test_folder(snippet)
     end
   end
 
+  def add_hosts()
+    Dir.chdir(__dir__) do
+      `sudo -- sh -c "cat ../hosts >> /etc/hosts"`
```
well1791: The problem when running only `sudo cat ../hosts >> /etc/hosts` is that `sudo` will scope only `cat` but the pipe to redirect output `>>` will be out of `sudo`, so basically `sudo` will prompt the user password to run just `cat`. Scoping everything into a single command shell make this work.

tools/snippet-testing/snippet_tester.rb
```diff
@@ -61,6 +61,16 @@ def test_folder(snippet)
...
+    end
+  end
+
+  def remove_hosts()
+    `sudo -- sh -c "sed -i '' '/^# START: api-snippets$/,/^# END: api-snippets$/d' /etc/hosts"`
```
well1791: Thanks for your observation. Actually, there is an error regarding mac `sed` and linux `sed` both are incompatible when using `-i` flag.
I already worked on this and I'll keep the linux version, and updated the readme to include a solution for mac users.
Regarding the command, you can see it like

```
sed
  -i'' # option and argument, notice there's no space! that's how it works in linux
  '/regex/,/regex/d' # regex range
  pathfile # the file sed will read
```

tools/snippet-testing/snippet_tester.rb
```diff
@@ -218,6 +228,10 @@ def parse_options(args)
       options.source_folder = File.expand_path(dir)
       options.test_default = true if options.test_default.nil?
     end
+
+    opts.on('-h', 'Append the twilio endpoints to your hosts file for testing') do
```
well1791: Good catch, I was doubting about it too.
I did a few changes and now I'm using `-u` flag, does that sound good to you?

well1791:
This feature won't be needed anymore because of Docker

PR #461, opened 2017-08-03
Adding prettier for js plus ci automation

well1791:
Refer to: #444


dprothero:
Wooooooo! 👍 Thanks for kicking this off @dkundel!

PR #464, opened 2017-08-08
Add examples for getting recordings for a given call

joliveros:


rest/recording/list-get-example-5/list-get-example-5.6.x.py
```diff
@@ -0,0 +1,15 @@
...
+from datetime import date
+
+# Your Account Sid and Auth Token from twilio.com/user/account
+account_sid = "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
+auth_token = "your_auth_token"
+client = Client(account_sid, auth_token)
+
+call_sid = "CAXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
+recordings = client.recordings \
+                   .list(call_sid=call_sid)
```
well1791: @joliveros PEP8 usually complains about this extra spaces as _over indented_, would you mind to just use 4 spaces instead of trying to align this to the `.`(dot) above? ^^

PR #465, opened 2017-08-08
Add python linter

well1791:


guides/request-validation-django/example-3/example-3.5.x.py
```diff
@@ -8,22 +8,24 @@
...
 
         # Validate the request using its URL, POST data,
         # and X-TWILIO-SIGNATURE header
-        request_valid = validator.validate(
-            request.build_absolute_uri(),
-            request.POST,
-            request.META.get('HTTP_X_TWILIO_SIGNATURE', ''))
+        request_valid = validator.validate(request.build_absolute_uri(),
+                                           request.POST,
+                                           request.META.get(
```
joliveros: Instead of this:
```python
request_valid = validator.validate(request.build_absolute_uri(),
                                                      request.build_absolute_uri(),
                                                      request.POST,
                                                      request.META.get(
                                                         'HTTP_X_TWILIO_SIGNATURE', ''))
```

Would something like this be better?  But probably not something the code fixer will do. May we can do another pull request later where we fix these types of things?
```python
twilio_sig = request.META.get('HTTP_X_TWILIO_SIGNATURE')

request_valid = validator.validate(request.build_absolute_uri(),
                                                      request.build_absolute_uri(),
                                                      request.POST,
                                                      twilio_sig)
```

well1791: Indeed! I also tried to fix some snippets that were like _bad_ formatted, but I think it's better to have a different PR to handle those things.

.flake8
```diff
@@ -0,0 +1,26 @@
...
+    # H401: docstring should not start with a space
+    H401,
+    # H403: multi line docstrings should end on a new line
+    H403,
+    # H404: multi line docstring should start without a leading new line
+    H404,
+    # H405: multi line docstring summary not separated with an empty line
+    H405,
+    # H501: Do not use self.__dict__ for string formatting
+    H501
```
smendes: I think that some of those shouldn't be ignored but fixed if possible. Namely H101 H202 H40x H501

README.md
```diff
@@ -170,6 +170,7 @@ Make sure you have the following dependencies installed on your system.
    export NVM_DIR="$HOME/.nvm" \
    [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" \
    nvm install --lts
+   npm install
```
smendes: Do we really need to run npm twice ?

add-ons/lookups/payfone-tcpa-compliance/payfone-tcpa-compliance.6.x.py
```diff
@@ -1,15 +1,16 @@
...
 
 number = client.lookups \
-               .phone_numbers("+16502530000") \
-               .fetch(add_ons="payfone_tcpa_compliance",
-                      add_ons_data={"payfone_tcpa_compliance.RightPartyContactedDate": "20160101"})
+    .phone_numbers("+16502530000") \
+    .fetch(
+        add_ons="payfone_tcpa_compliance",
+        add_ons_data={
+            "payfone_tcpa_compliance.RightPartyContactedDate": "20160101"})
```
smendes: We should tweak yapf to have closing brackets and parentheses to match the indentation.

guides/request-validation-python/example-2/example-2.5.x.py
```diff
@@ -1,9 +1,10 @@
 from flask import Flask, request
 from twilio import twiml
-
+from . import validate_twilio_request
```
smendes: Relative import in Python works only inside packages (directories containing a `__init__.py` file). For snippets we shouldn't rely on any other local files.

install_travis_dependencies.sh
```diff
@@ -16,6 +16,7 @@ sudo apt-get install -y --force-yes gradle
 if [[ $languages == *"python"* ]]; then
   sudo apt-get install -y --force-yes python-dev python-pip
   sudo pip install --upgrade pip wheel virtualenv virtualenvwrapper --quiet
+  sudo pip install -r requirements.txt
```
smendes: This installs virtualenv and then install the dependencies without using it. Travis spawn a new instance every time, we shouldn't need to use a package version manager.

rest/taskrouter/twiml/example4/example/example.5.x.py
```diff
@@ -1,13 +1,18 @@
 # Download the Python helper library from twilio.com/docs/python/install
-from flask import Flask, request, Response
-import twilio.twiml
+from flask import Flask
+import twilio.twimni
```
smendes: Typo

README.md
well1791: I don't see where is `npm` duplicated

guides/request-validation-python/example-2/example-2.5.x.py
well1791: Actually, this folder `guides/` is not tested at all, so right now I'm thinking it's better to revert its changes, and exclude this folder from the linting.
wdyt?

install_travis_dependencies.sh
well1791: Actually, the virtual env wrapper is used by the test runner, what I'm doing here is installing dependencies globally and using it before the test runner starts with its job (which depends on the virtual env).
Also, I don't think I should depend on a virtual env as I'm linting files `5.x.py` and `6.x.py` (meaning all `.py` files), there's no need to create an env for something globally.

add-ons/lookups/payfone-tcpa-compliance/payfone-tcpa-compliance.6.x.py
well1791: Going to check that 👍

well1791: By default, it sets this property `align_closing_bracket_with_visual_indent` to `True`

rest/queue/list-post-example-1/list-post-example-1.5.x.py
```diff
@@ -3,8 +3,10 @@
...
 account_sid = "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
-auth_token  = "your_auth_token"
+auth_token = "your_auth_token"
 client = TwilioRestClient(account_sid, auth_token)
 
-queue = client.queues.create("newqueue", )
-print(queue.sid)
\ No newline at end of file
+queue = client.queues.create(
+    "newqueue",
```
joliveros: I think this comma at the end is causing the code fixer to wrap the arguments

rest/recording/get-recording-xml/get-recording-xml.6.x.py
```diff
@@ -4,8 +4,9 @@
 # Your Account Sid and Auth Token from twilio.com/user/account
 account_sid = "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
 auth_token = "your_auth_token"
-recording_sid="RE557ce644e5ab84fa21cc21112e22c485" 
+recording_sid = "RE557ce644e5ab84fa21cc21112e22c485"
 
 client = Client(account_sid, auth_token)
 
-client.api.v2010.accounts(sid=account_sid).recordings(sid=recording_sid).fetch()
+client.api.v2010.accounts(sid=account_sid).recordings(sid=recording_sid
```
joliveros: is there a way to configure the linter to wrap the code this way instead?
```
client.api.v2010.accounts(sid=account_sid)\
     .recordings(sid=recording_sid).fetch()
```

rest/queue/list-post-example-1/list-post-example-1.5.x.py
well1791: Good catch! actually, that can be the reason! I'll try some changes and see if the formatter does the same thing.

rest/recording/get-recording-xml/get-recording-xml.6.x.py
well1791: I would like to, but I haven't found any option that handles this, actually, the problem is when the line is over 80 chars, but yapf doesn't look like solving it (regarding the indentation)

PR #469, opened 2017-08-09
Add create-address snippets for old versions

well1791:


rest/addresses/instance-create-example/instance-create-example.6.x.java
```diff
@@ -0,0 +1,29 @@
...
+import com.twilio.sdk.resource.instance.Address;
+import java.util.HashMap;
+
+public class Example {
+
+  // Find your Account Sid and Token at twilio.com/user/account
+  public static final String ACCOUNT_SID = "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX";
+  public static final String AUTH_TOKEN = "your_auth_token";
+
+  public static void main(String[] args) throws TwilioRestException {
```
hortega: In Java snippets we use 4 space tabs.

hortega:
I don't know exactly what the story is about, but `instance-create-example` and `list-post-example-1` are very similar, and I think `list-post-example-1` is not referenced in the CMS, are you sure we need this new code sample?

PR #477, opened 2017-08-14
Add snippets for deployed devices, Node and Curl

dr0nius:
Add snippets for Fleets, Deployments, Keys, Certificates, Devices


deployed-devices/rest/keys/create-key/output/create-key.json
```diff
@@ -0,0 +1,11 @@
+{
+  "sid": "KYXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
+  "friendly_name": null,
+  "fleet_sid": "FLXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
+  "account_sid": "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
+  "device_sid": "THXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
+  "secret": "abcdeghijklmnopqrstuvwxyz",
```
coderkat: nitpick: for secrets, we like to stick to a placeholder that is more clearly a needs-to-be-replaced placeholder, like 'your_secret'

deployed-devices/rest/certificates/create-certificate/output/create-certificate.json
```diff
@@ -0,0 +1,11 @@
+{
+  "sid": "CYXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
+  "friendly_name": null,
+  "fleet_sid": "FLXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
+  "account_sid": "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
+  "device_sid": "THXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
+  "thumbprint": "00112233445566778899aabbccddeeff00112233445566778899aabbccddeeff",
```
coderkat: Is this thumbprint a static string? Something users know or generate? If the string here is a placeholder for something variable, it might be good to use a string like 'your_generated_thumbprint' instead. If not, carry on! 

(there are a few other places this occurs - if you do change this string, just double-check that it's the same across the board. thanks!)

dr0nius: This string is SHA256 over the supplied certificate data, so it will depend on user input. In theory, they could know upfront what it is, but not mandatory. Happy to rename to <supplied_certificate_fingerprint>

deployed-devices/rest/keys/create-key/output/create-key.json
dr0nius: Aye, fixing.

PR #495, opened 2017-09-05
DEVED-2520: Update Node.js validationRequests.create

imthepitts:


atbaker:
@imthepitts It seems our build process is a little borked right now. Merging anyway.

imthepitts:
@atbaker looks like the problem was I updated the legacy version (need to revert), but there doesn't seem to be any next-gen in this list -- where is its source?

https://github.com/TwilioDevEd/api-snippets/tree/master/rest/outgoing-caller-ids/list-post-example-1

PR #505, opened 2017-09-15
double quote changed to single quote for prevent error

shibby:


ip-messaging/push-config/new-message-push-config/new-message-push-config.php
```diff
@@ -19,6 +19,6 @@
 $service = $client->chat->v2->services($CHAT_SERVICE_SID);
 $response = $service->update(array(
     "Notifications.NewMessage.Enabled" => "true",
-    "Notifications.NewMessage.Template" => "A New message in ${CHANNEL} from ${USER}: ${MESSAGE}",
+    "Notifications.NewMessage.Template" => 'A New message in ${CHANNEL} from ${USER}: ${MESSAGE}',
```
well1791: @shibby 

I think those double quotes `"` are supposed to work with string interpolation in PHP, the current code is trying to evaluate variables `$CHANNEL`, `$USER` and `$MESSAGE` which are not defined anywhere so it's definitely an error.
But, if it's supposed to evaluate those variables, I think more natural to write that interpolation like this:

```php
 "Notifications.NewMessage.Template" => "A New message in {$CHANNEL} from {$USER}: {$MESSAGE}",
```

Else, if the value for the `Notifications.NewMessage.Template` key is supposed to be a string template, perhaps it's more convenient to write it like:

```php
"Notifications.NewMessage.Template" => 'A New message in {CHANNEL} from {USER}: {MESSAGE}',
```

shibby:
hi @well1791 thank you for review. 

`${CHANNEL}`, `${USER}` and `${MESSAGE}` are variables for twilio's push notification template. Here is default value of notification text:

screenshot from 2017-09-16 14-09-26

So, Twilio's example is wrong since it will throw error. 
We can't use your second suggestion since Twilio's variable are known as above. 
The only way is, using single quote. 

I hope i don't miss any piece :)

PR #508, opened 2017-09-18
WIP - Docker image to run tests

joliveros:


Makefile
```diff
@@ -0,0 +1,67 @@
...
+127.0.0.1 lookups.twilio.com
+127.0.0.1 messaging.twilio.com
+127.0.0.1 monitor.twilio.com
+127.0.0.1 notifications.twilio.com
+127.0.0.1 notify.twilio.com
+127.0.0.1 preview.twilio.com
+127.0.0.1 pricing.twilio.com
+127.0.0.1 sync.twilio.com
+127.0.0.1 taskrouter.twilio.com
+127.0.0.1 video.twilio.com
```
well1791: @joliveros is there a way to move these paths to a file and read the content of the file instead of defining it inside the Makefile?
In case someone needs to add a new path, the commit will show the git-diff within the file `hosts` (assuming that's the name of the file) and that's clear to read ^^

joliveros: that makes sense. I'll see if I can figure it out :)

PR #510, opened 2017-09-19
Demonstrate common additive permission updates

well1791:


ip-messaging/rest/roles/update-role/update-role.3.x.js
```diff
@@ -5,10 +5,16 @@ const Twilio = require('twilio').Twilio;
 const client = new Twilio(accountSid, authToken);
 const service = client.chat.services('ISXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX');
 
+role = service.roles('RLXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX').fetch
```
joliveros: it seems these lines are ruby?

PR #520, opened 2017-09-27
Add examples for answering machine detection

joliveros:


rest/answering-machine-detection/outgoing-call/outgoing-call-1.6.x.py
```diff
@@ -0,0 +1,13 @@
...
+from twilio.rest import Client
+
+# Your Account Sid and Auth Token from twilio.com/user/account
+account_sid = "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
+auth_token = "your_auth_token"
+client = Client(account_sid, auth_token)
+
+call = client.calls("CA42ed11f93dc08b952027ffbc406d0868")\
+    .update(to='+1562300000',
+            from='+18180000000',
```
acamino: @joliveros, would you mind updating this snippet? `from` is a reserved keyword.

acamino:
@joliveros, would you mind checking the errors you get on CI for Ruby? The build is failing because `rubocop` is complaining about the usage of double quotes instead of single quotes. https://travis-ci.org/TwilioDevEd/api-snippets/jobs/280486387

I would recommend running _rubocop_ and similar tools. What do you think?

rest/answering-machine-detection/outgoing-call/outgoing-call-1.5.x.rb
```diff
@@ -0,0 +1,19 @@
+# Get twilio-ruby from twilio.com/docs/ruby/install
+require 'twilio-ruby'
+
+# Get your Account SID and Auth Token from twilio.com/console
+account_sid = 'ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'
+auth_token = 'your_auth_token'
+
+# Initialize Twilio Client
+@client = Twilio::REST::Client.new(account_sid, auth_token)
```
acamino: @joliveros, this is a subtle issue that we might want to discuss. `@` is used for instance variables. Most of the examples we have here are not using a class, so making them instance variables is unnecessary.

The following contrived example tries to illustrate my point.
```ruby
class Example
  def initialize(client)
    @client = client
  # ^ In this example, the usage of an instance variable makes sense.
  end

  def make_call
    call = @client.api.calls('CA42ed11f93dc08b952027ffbc406d0868')
    # ...
  end
end
```

What do you think about removing the prefix `@` from `@client` for the new snippets starting now?

joliveros: I agree. I plead guilty to copy and paste :smile:

rest/answering-machine-detection/outgoing-call/outgoing-call-1.7.x.java
```diff
@@ -0,0 +1,19 @@
...
+public class Example {
+  // Find your Account Sid and Token at twilio.com/user/account
+  public static final String ACCOUNT_SID = "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX";
+  public static final String AUTH_TOKEN = "your_auth_token";
+
+  public static void main(String[] args) {
+    Twilio.init(ACCOUNT_SID, AUTH_TOKEN);
+
+    Call call = CallUpdater("CA42ed11f93dc08b952027ffbc406d0868")
+        .setTo('+1562300000')
```
acamino: I'm pretty sure this example won't work since strings should be enclosed in quotes. Literal characters use single quotes.

rest/answering-machine-detection/outgoing-call/outgoing-call-1.6.x.py
```diff
@@ -0,0 +1,14 @@
+# Download the Python helper library from twilio.com/docs/python/install
+from twilio.rest import Client
+
+# Your Account Sid and Auth Token from twilio.com/user/account
+account_sid = "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
```
acamino: I'd like to keep the usage of quotes consistent across the snippets. Perhaps you want to take a look at https://stackoverflow.com/questions/56011/single-quotes-vs-double-quotes-in-python. Moreover, I want to take a look at the coding conventions we're using for Python in this particular case.

rest/answering-machine-detection/outgoing-call/outgoing-call-1.5.x.rb
acamino: Thank you!

rest/answering-machine-detection/outgoing-call/outgoing-call-1.5.x.cs
```diff
@@ -0,0 +1,25 @@
...
+    static void Main(string[] args)
+    {
+        // Find your Account Sid and Auth Token at twilio.com/console
+        const string accountSid = "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX";
+        const string authToken = "your_auth_token";
+        TwilioClient.Init(accountSid, authToken);
+
+        var call = CallResource.Update(
+            "CA42ed11f93dc08b952027ffbc406d0868",
+            to: '+1562300000',
```
acamino: A literal _String_ must be enclosed in double quotes. Could you please change this snippet to use an appropriate syntax?

joliveros: good catch!

PR #524, opened 2017-10-02
Notify uses Objects instead of Strings now in Notification resource

devidliik:
This PR aligns Java and Node helper libs with latest Twilio SDK release.


dprothero:
LGTM

PR #528, opened 2017-10-06
Add examples for request validation using ASP.NET Web API

acamino:
These examples will be used in the tutorial for **Secure your C# / ASP.NET app by validating incoming Twilio requests**.

It contains the actual end-points:
- /voice
- /message

And the filter as a custom data annotation.

### DIFF
--- a/guides/request-validation-csharp-web-api/example-2/example-2.4.x.cs
+++ b/guides/request-validation-csharp-web-api/example-2/example-2.4.x.cs
@@ -1,0 +1,62 @@
+using System.Net.Http;
+using System.Text;
+using System.Web.Http;
+using Twilio.TwiML;
+using ValidateRequestExample.Filters;
+
+namespace ValidateRequestExample.Controllers
+{
+    public class TwilioMessagingRequest
+    {
+        public string Body { get; set; }
+    }
+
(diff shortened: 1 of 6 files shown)


guides/request-validation-csharp-web-api/example-2/example-2.4.x.cs
```diff
@@ -0,0 +1,61 @@
...
+    public class IncomingController : ApiController
+    {
+        [Route("voice")]
+        [AcceptVerbs("POST")]
+        [ValidateTwilioRequest]
+        public HttpResponseMessage PostVoice([FromBody] TwilioVoiceRequest voiceRequest)
+        {
+            var message =
+                "Thanks for calling! " +
+                $"Your phone number is {voiceRequest.From}. I got your call because of Twilio's webhook. " +
```
well1791: @acamino can you split this line to make it more readable

```cs
$"Your phone number is {voiceRequest.From}. " +
"I got your call because of Twilio's webhook. " +
```

acamino: Thanks for the suggestion @joliveros!

PR #537, opened 2017-10-17
Update all C# snippets for twilio-csharp v5.8

joliveros:


twiml/message/message/message-3/message-3.5.x.cs
```diff
@@ -7,8 +8,8 @@ static void Main()
     {
         var response = new MessagingResponse();
         response.Message("Store Location: 123 Easy St.",
-            action: "/SmsHandler.php", method: "POST");
+            action: new Uri("/SmsHandler.php"), method: "POST");
```
well1791: @joliveros I think this `method: "POST"` should be `method: HttpMethod.Post`

twiml/voice/gather/gather-3/gather-3.5.x.cs
```diff
@@ -1,17 +1,19 @@
...
 class Example
 {
     static void Main()
     {
         var response = new VoiceResponse();
-        var gather = new Gather(action: "/process_gather.php", method: "GET");
+        var gather = new Gather(action: new Uri("/process_gather.php"), method: "GET");
         gather
             .Say("Please enter your account number,\nfollowed by the pound sign");
         response.Gather(gather);
```
well1791: @joliveros Also, this change should be `response.Append(gather)`

twiml/message/message/message-3/message-3.5.x.cs
joliveros: In the tests for twilio-csharp they just use a string. wdyt?
https://github.com/twilio/twilio-csharp/blob/master/test/Twilio.Test/TwiML/VoiceResponseTest.cs#L25

twiml/voice/gather/gather-3/gather-3.5.x.cs
joliveros: good catch

joliveros:
@well1791, I think fixed the issues you mentioned. Mind having another look?

twiml/message/message/message-3/message-3.5.x.cs
well1791: It will work, but it's preferable to use what they provide as they may change the `method` parameter to depend on a type interface (like [here](https://github.com/twilio/twilio-csharp/blob/master/src/Twilio/Rest/Taskrouter/V1/Workspace/Worker/ReservationOptions.cs#L215)). Also, it's less prone to write typos when using the language syntax check instead of writing it manually.

joliveros: sounds like a plan. I'll update that now

PR #545, opened 2017-10-22
fix paho-mqtt usage in python mqtt quickstart

Brodan:
Inside the [python-mqtt-quickstart](https://www.twilio.com/docs/quickstart/sync-iot/python-mqtt-quickstart), the Python snippet for the [subscribe-to-the-sync-document-from-mqtt](https://www.twilio.com/docs/quickstart/sync-iot/python-mqtt-quickstart#subscribe-to-the-sync-document-from-mqtt) section has an error in it. I noticed it after pasting it and running it locally:
```
Traceback (most recent call last):
  File "mqtt.py", line 10, in <module>
    client.set_tls(None, pem_location, key_location)
AttributeError: 'Client' object has no attribute 'set_tls'
```
As per the [paho-mqq docs](https://pypi.python.org/pypi/paho-mqtt/1.1#option-functions), `set_tls()` should be `tls_set()`.

Additionally, the [subscribe-and-print.py](https://github.com/TwilioDevEd/api-snippets/blob/master/deployed-devices/quickstarts/paho-python/subscribe-and-print.py) subscribes to the `'MyDoc'` Document, while the rest of the quickstart uses `'MyFirstDocument'`, so the final step of the quickstart does not behave as intended. The `subscribe-and-print.py` snippet should be updated to use the same Document name as the other snippets.

The text in this quickstart should also be updated to match the above: `When you run this script, the current current state of MyDoc will be printed upon connection;` should use `MyFirstDocument` instead of `MyDoc`.


Brodan:
Also, checks are failing because of unrelated Ruby/C# stuff.

PR #551, opened 2017-11-15
Stats endpoints utilizing latest helper libs methods in a friendlier manner

jwitz10:


jwitz10:
Looks like the build is failing. This appears to be because the ruby and php versions are out of date for these helper lib updates... thus the methods that I'm referencing here do not exist.

Ruby needs to be upgraded from 5.3.0 to 5.3.1
Node needs to be upgraded from 3.7.0 to 3.8.0
PHP needs to be upgraded from 5.15.0 to 5.15.1
C# needs to upgraded from 5.7.0 to 5.7.1

Not sure if this is controlled at the travis-ci level or repo level.

jwitz10:
I was able to figure out the above.

PR #552, opened 2017-11-16
Remove date-created dates

coderkat:
These make Google think this page was published/last updated in 2012. Not good for building confidence in our docs


pkamp3:
The very heaviest of plus signs!

PR #557, opened 2017-11-22
Fix helper libraries taglines

smendes:


coderkat:
👍  amazing, thank you

PR #561, opened 2017-11-22
Proxy public beta update

pkamp3:
Updating Proxy for Beta


pkamp3:
These tests will break until libraries regenerate

PR #567, opened 2017-11-28
Sync TTL related changes to snippets

dr0nius:
- Add new root update methods to list, map and stream objects
- Add date_expires field to document, list, map and stream objects; also list/map items
- Add TTL parameter to document, list, map, stream creation as an example
- Add TTL parameter to list and map item creation
- Minor syntax tweaks for better consistency


dr0nius:
Note that I needed to bump all helper library versions in tests for compatibility reasons.
Also, this interface extension is required for HTTP mock: https://github.com/TwilioDevEd/twilio-api-faker/pull/10

PR #582, opened 2017-12-14
Notify - Adds instance id service sample

viktormuller:
Adds an instance id service sample to FCM registration guide


coderkat:
👍

PR #584, opened 2017-12-15
Swap names and some example text for gender/situational neutrality

jennifermarie:
Retires the 'Jenny' reference and replaces instances of Jenny/Tommy with more gender-neutral Joey/Charlie. Replaces Jenny/Tommy narrative with neutral and/or geeky references.

NB: the 867-5309 number is still in place because I didn't want to break anything and it's used fairly widely in this repo. If it's safe to swap for another number, we can do that as well.


jennifermarie:
If the reviewer on this PR could please tag me when you get to it, that would be amazing. There are corresponding changes in the prose text on many of our docs that need to happen as well, and I don't want to have these out of sync. Thanks!

jennifermarie:
Great, thanks @joliveros! New commit incoming, with numbers swapped (I'm keeping it simple -- 5309 -> 5310)

PR #586, opened 2017-12-21
Update capability token creation method

ayyrickay:
Old method was `generate()`, it's now `to_jwt()`


atbaker:
Looks like our build is failing, but I'm not seeing this snippet in the list of problematic ones (we've got some known issues we're working through). Merging!

PR #587, opened 2018-01-03
Fixed grammar issue

liusqi:
update('accept') will throw errors. I think it's because the reservation api has changed.


atbaker:
Appreciate the fix - thank you!

PR #593, opened 2018-01-23
Update params for successful signature validation

mjwang:
Signature validation fails. Caller and From params should be +14158675309 and not +14158675310

Should be changed for other languages as well.


po5i:
We updated the signature hash instead. Thanks.

PR #596, opened 2018-01-30
Adding missing parameters

jamesfernandes:
Bringing code into line with caption that says this example should _Only show completed calls started between midnight Jul 04th, 2009 and midnight Jul 06th, 2009._


po5i:
Fixes #597

PR #609, opened 2018-02-18
Fix: workspace.workers is a method call not a static property

RoyHP:


atbaker:
Appreciate the fix - thank you!

PR #614, opened 2018-02-21
DEVED-3119:  Participants /docs shows same code for 4 different cases

po5i:
Fixed snippets for node, python, ruby and php. They were all the same code for `retrieve-a-participant` snippet.


dprothero:
LGTM

PR #619, opened 2018-02-28
Updated 2FA Authy verify webhook with new signature verification

po5i:
Updated Python, Ruby, JavaScript and PHP snippets. Previously the Authy request had individual form parameters, now it comes with a JSON body, so it changed the way they have to be read and sorted.

@hortega @joliveros could you give me a hand with Java and C# snippets when you can, please?

### DIFF
--- a/two-factor-authentication/verify-webhook/verify-webhook.5.x.rb
+++ b/two-factor-authentication/verify-webhook/verify-webhook.5.x.rb
@@ -8,16 +8,9 @@ class CallbackVerifier
   def verify_callback(request, api_key)
     url = url_for(:only_path => false, :overwrite_params=>nil)
 
-    # Sort and join the parameters
-    parameter_string = request
-      .query_parameters
-      .to_a
-      .concat(request.request_parameters.to_a)
-      .sort
-      .map{ |p| p.join('=')}
-      .join('&')
-
-    parameter_string = URI.encode(parameter_string)
+    # Sort and join the parameters on Rails
+    json_params = JSON.parse(request.body.read)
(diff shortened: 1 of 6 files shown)


two-factor-authentication/verify-webhook/verify-webhook.5.x.rb
```diff
@@ -8,16 +8,9 @@ class CallbackVerifier
...
-      .to_a
-      .concat(request.request_parameters.to_a)
-      .sort
-      .map{ |p| p.join('=')}
-      .join('&')
-
-    parameter_string = URI.encode(parameter_string)
+    # Sort and join the parameters on Rails
+    json_params = JSON.parse(request.body.read)
+    parameter_string = json_params.to_query
```
hortega: Are the parameters in the string sorted in alphabetical order?

po5i: Good catch. I don't see the sort anywhere but it "magically" worked when I tested it with the Authy JSON body.
I will review it more in detail.

po5i: It looks like on Rails `Hash#to_query` extension [sorts the query parameters](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/object/to_query.rb#L82) before joining them.

hortega: Nice one! Gotta love magic :slightly_smiling_face:

coderkat:
👍

PR #623, opened 2018-03-07
bugfix: pass RECORDING_ID instead of AUTH_TOKEN

innamhunzai:
Recording fetcher expects parameter 2 to  be recording_id, auth_token is passed instead.


atbaker:
Appreciate the fix! Thank you!

PR #629, opened 2018-03-21
lookups code samples should demonstrate catching the 404 exception

po5i:
Lookups API does something that other APIs don't do – it returns an HTTP 404 when it can't find the carrier info for a number, which results in a runtime exception in helper libraries.

https://www.twilio.com/docs/api/lookups

Each code sample (for every language) on this page should demonstrate how to catch the runtime exception and check for the 404.


coderkat:
👍

PR #633, opened 2018-04-09
Update send_notifications.6.x.py

FahadAlbukhari:


po5i:
Thanks for your PR. This is not a typo on the snippet as you can see that in the [library's source code](https://github.com/twilio/twilio-python/blob/master/twilio/rest/api/v2010/account/message/__init__.py#L72).

PR #635, opened 2018-04-19
Update tokens in sig verfication code to 12345

dxm7:


dxm7:
Tokens actually verify on secret of 12345, see other code samples in py and java in that folder. csharp devs wrote in that our sample code doesn't verify and noticed that they should all be set to 12345.

po5i:
Makes sense! Thanks.

PR #639, opened 2018-04-21
Prettier ruby code

Dorian:
@po5i


po5i:
Thanks for doing this @Dorian 

I have only one question. Why did you remove the `require 'rubygems'` lines, I know they useless in mainstream versions of Ruby, but don't you think it would be nice to have for legacy reasons?

Dorian:
> Ruby 1.9 and newer ships with RubyGems built-in

http://guides.rubygems.org/rubygems-basics/

@po5i And most people are on 1.9+ (1.8.2 is 14 years old :) )

PR #663, opened 2018-07-05
Translate "recording transcriptions" code snippet to helper lib languages

po5i:
Adds sample code for fetching the transcriptions list from a given recording.

This is the continuation for #659


coderkat:
👍 Thanks @po5i !

Issue #673, opened 2018-07-31
Validate Twilio Request Attribute raw url


jmoceri34:
Referring to this [code snippet](https://github.com/TwilioDevEd/api-snippets/blob/master/guides/request-validation-csharp/example-1/example-1.cs), line 34

```csharp
            var requestUrl = request.RawUrl;
```
was giving me everything following the host (www.google.com/x/y/z, only returning /x/y/z)

I changed to

```csharp
var requestUrl = request.Url.AbsoluteUri;
```

in order to get the validation to work

Using nuget packages 

Microsoft.AspNet.Mvc 5.2.3
Twilio 5.14.0
Twilio.Aspnet.Mvc 5.9.7

PR #674, opened 2018-08-01
Change RawUrl to AbsoluteUri

dxm7:
RawUrl only returns content after the url following the domain information. The request calculation should be based on the full Url which AbsoluteUri returns.


po5i:
Closes #673.

Thank you!

PR #685, opened 2018-09-02
Fix MessagingResponse import statement

duffn:


guides/request-validation-python/example-2/example-2.6.x.py
```diff
@@ -1,5 +1,6 @@
 from flask import Flask, request
-from twilio.twiml.voice_response import VoiceResponse, MessagingResponse
+from twilio.twiml.messaging_response import MessagingResponse
+from twilio.twiml.voice_response import VoiceResponse
```
georgemandis, line 3: 👍

PR #708, opened 2018-10-23
Adding snippets for TaskRouter TaskChannel create, update, retrieve, and delete

alexdlaird:


rest/taskrouter/taskchannels/list/post/example-1/output/example-1.json
```diff
@@ -0,0 +1,10 @@
+{
+  "sid": "TC72e66c2c822002ff3cad36c608979ee9",
+  "account_sid": "AC1c41896f44563b7c8194235ee6d267b1",
```
joliveros: Maybe a good idea to obfuscate the real account_sid etc?

alexdlaird: I assume this was a non-function SID, since it's used in other examples too. But no, it's tied to an account 🙈. I'll go through and obfuscate any other uses of it as well. Thanks.

PR #715, opened 2018-11-14
added m2m_commands_feather32u4fona

cskonopka:
I added a new Quickstart to Wireless:

m2m_commands_feather32u4fona
- This is an example of how to create M2M commands using the Adafruit Feather 32u4 FONA board. The code follows the same paradigm as the Arduino/SIM800 example.

m2m_commands_arduino_sim800
- Added this back since I messed up the CMS yesterday. Good to go.


pkamp3:
Lookin' good

PR #722, opened 2018-11-20
Fix python example snippets

patillacode:
File `example.5.x.py` seems fine (tested under twilio==5.7.0)
Fixed `example.6.x.py` (tested under twilio==6.20.0)

Also fixed possible issue with XML syntax in output example, unexpected `/`. 
(not sure if out of the task's scope)

### DIFF
--- a/rest/taskrouter/twiml/example2/example/example.4.x.php
+++ b/rest/taskrouter/twiml/example2/example/example.4.x.php
@@ -1,6 +1,6 @@
 <?php
 // Download the library and copy into the folder containing this file.
-require('/path/to/vendor/twilio/sdk/Services/Twilio.php');
+require_once '/path/to/vendor/twilio/sdk/Services/Twilio.php';
 
 $response = new Services_Twilio_Twiml;
 $response->enqueue(array('workflowSid' => 'WW0123456789abcdef0123456789abcdef'))
--- a/rest/taskrouter/twiml/example2/example/example.4.x.php
+++ b/rest/taskrouter/twiml/example2/example/example.4.x.php
@@ -1,13 +1,12 @@
 <?php
 // Download the library and copy into the folder containing this file.
-require('twilio-php/Services/Twilio.php');
+require('/pat/to/vendor/twilio/sdk/Services/Twilio.php');
 
 $response = new Services_Twilio_Twiml;
-$response->enqueue(array(workflowSid => 'WW0123456789abcdef0123456789abcdef'))
-	->task("{'account_number':'12345abcdef'}");
+$response->enqueue(array('workflowSid' => 'WW0123456789abcdef0123456789abcdef'))
+    ->task("{'account_number':'12345abcdef'}");
 print $response;
 ?>
-
 <!-- alternatively -->
 
 <?php
--- a/rest/taskrouter/twiml/example2/example/example.4.x.php
+++ b/rest/taskrouter/twiml/example2/example/example.4.x.php
@@ -1,6 +1,6 @@
 <?php
 // Download the library and copy into the folder containing this file.
-require('/pat/to/vendor/twilio/sdk/Services/Twilio.php');
+require('/path/to/vendor/twilio/sdk/Services/Twilio.php');
 
 $response = new Services_Twilio_Twiml;
 $response->enqueue(array('workflowSid' => 'WW0123456789abcdef0123456789abcdef'))
--- a/rest/taskrouter/twiml/example2/example/example.5.x.cs
+++ b/rest/taskrouter/twiml/example2/example/example.5.x.cs
@@ -4,21 +4,69 @@ using System;
 using System.Net;
 using Twilio.TwiML;
 
-class Example
-{
-    public static HttpListenerResponse SendResponse(HttpListenerContext ctx)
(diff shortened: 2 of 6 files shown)


patillacode:
- JS example files seem fine 
- PHP `5.x `version example file seema fine (tested under` twilio/sdk 5.25.0`)
- PHP `4.x` version example file: Minor changes (php warning & better lib path) tested under `twilio/sdk 4.12.1`

rest/taskrouter/twiml/example2/example/example.4.x.php
```diff
@@ -1,13 +1,12 @@
 <?php
 // Download the library and copy into the folder containing this file.
-require('twilio-php/Services/Twilio.php');
+require('/pat/to/vendor/twilio/sdk/Services/Twilio.php');
```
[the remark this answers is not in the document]
patillacode: Fixed! Thanks!

patillacode:
- C# 5.x version example has been edited (tested under `Twilio 5.22.0`)

Notes to consider:

- I saw that the pattern in the examples is to have some sort of a "run this and it works" so I added a `main` method that calls the `SendResponse` method.

- In the rest of the snippets the webserver is somehow embedded  within the file, so in order to follow the same pattern I added a HTTP Server within the `main` method. 

- It might seem like an overkill for an experienced C# developer but this has been my very first contact with C#, I had many issues just to get the environment right and I just wanted the snippet to follow the same pattern.

- I did try to use `SimpleWebServer` (as I saw in some Twilio examples) but I couldn't find the package I need to install in order to use it so I kind of went rogue and explicit as you can see.

rest/taskrouter/twiml/example2/example/example.5.x.cs
```diff
@@ -4,21 +4,69 @@
...
+        {
+            HttpListenerRequest request = ctx.Request;
+            HttpListenerResponse response = ctx.Response;
+
+            response.StatusCode = (int)HttpStatusCode.OK;
+            response.ContentType = "text/xml";
+
+            var twiml = new VoiceResponse();
+            twiml.Enqueue("{\"account_number\":\"12345abcdef\"}",
+                workflowSid: "WW0123456789abcdef0123456789abcdef");
```
maylonpedroso: This code here is generating this TwiML
```
<Response>
  <Enqueue workflowSid="WW0123456789abcdef0123456789abcdef">{"account_number":"12345abcdef"}</Enqueue>
</Response>
```
and it doesn't match the expected output

rest/taskrouter/twiml/example2/example/example.4.x.php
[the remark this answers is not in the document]
patillacode: I have added `require_once` but I left the path since it is version `4.x`
Sounds good?

rest/taskrouter/twiml/example2/example/example.5.x.cs
patillacode: Thanks @maylonpedroso 
I have edited it now.

If you are happy with this change I will remove all the "web server overkill"

patillacode: I forgot to mention @maylonpedroso 

I think `response.StatusDescription = twiml.ToString();` could be a problem, I did get the error below when testing with that line:

```
Unhandled Exception: System.ArgumentException: Specified value has invalid Control characters.
Parameter name: name
   at System.Net.HttpListenerResponse.set_StatusDescription(String value)
   at project.TwilioExample.SendResponse(HttpListenerContext ctx)
```

This is why I decided to return two values, so I could output the xml properly in the server.

Any thoughts on this?

patillacode:
- Java `7.x` version example has been edited 
(tested under `twilio-7.31.0` and `twilio-java-sdk-7.0.0-rc-10`)

rest/taskrouter/twiml/example2/example/example.5.x.cs
maylonpedroso: Hi @patillacode,
The exception is because StatusResponse does not support `\n` char in the content. What do you think about doing this:
```
public static void SendResponse(HttpListenerContext ctx)
{
    HttpListenerRequest request = ctx.Request;
    HttpListenerResponse response = ctx.Response;

    response.StatusCode = (int)HttpStatusCode.OK;
    response.ContentType = "text/xml";

    // Create TwiML
    var twiml = new VoiceResponse();
    var enqueue = new Enqueue(workflowSid: "WW0123456789abcdef0123456789abcdef");
    enqueue.Task("{\"account_number\": \"12345abcdef\"}");
    twiml.Append(enqueue);

    // Write the output and close the stream
    byte[] buffer = Encoding.UTF8.GetBytes(twiml.ToString());
    response.ContentLength64 = buffer.Length;
    response.OutputStream.Write(buffer,0, buffer.Length);
    response.OutputStream.Close();
}
```

patillacode: Hello there @maylonpedroso 

Sounds good to me, thanks, I was confused on how much of the _http server_ I could keep in the snippet.

Adding this. Pushing shortly.

PR #730, opened 2018-12-21
Composition hooks

lwilio:


video/rest/compositionhooks/list-hooks/meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "title": "Create a Composition Hook mixing all room audio tracks",
```
rtvlad: The title is wrong, should be `List all enabled Composition Hooks of an account`

video/rest/compositionhooks/grid-mixing-hook/meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "title": "List all enabled Composition Hooks",
```
rtvlad: Interchange this title with the one from list composition hooks.

PR #741, opened 2019-01-31
Updates signatures for validation snippets.

philnash:
While replacing test numbers, the signatures weren't updated so the validation snippets were all invalid.

Also, Ruby prefers `puts` to `print`.


dprothero:
:+1:

PR #747, opened 2019-02-23
Use `to_jwt` method instead of erroneous `generate` call

craigsdennis:
`generate` is not a method of ClientCapabilityToken or the base class JWT.


craigsdennis:
I manually verified that this works for the Twilio.js quick start

PR #748, opened 2019-03-04
Adding snippets for Chat token management

dr0nius:
Particularly addressing three scenarios:
- initializing the client with initial token
- updating the token with provided string
- using the built-in token lifecycle events

### DIFF
--- a/ip-messaging/tokens/initialization/initialization.java
+++ b/ip-messaging/tokens/initialization/initialization.java
@@ -1,0 +1,14 @@
+import com.twilio.chat.ChatClient;
+import com.twilio.chat.ChatClientListener;
+
+ChatClient.Properties.Builder builder = new ChatClient.Properties.Builder();
+ChatClient.Properties properties = builder.createProperties();
+// See below for event listener implementation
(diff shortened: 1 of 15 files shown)


ip-messaging/tokens/initialization/initialization.java
```diff
@@ -0,0 +1,14 @@
+import com.twilio.chat.ChatClient;
+import com.twilio.chat.ChatClientListener;
+
+ChatClient.Properties.Builder builder = new ChatClient.Properties.Builder();
+ChatClient.Properties properties = builder.createProperties();
+// See below for event listener implementation
+ChatClientListener listener = new ChatClientListener();
```
berkus: `ChatClientListener` is an interface, you can't create an instance.

berkus: Use `Listener` probably, as this is what is used below in the examples.

dr0nius: Ouch, facepalm

berkus:
👍

PR #750, opened 2019-03-06
Environment variable examples

dprothero:
Examples in supported languages for reading environment variables.


craigsdennis:
Yay!

craigsdennis:
I know the Python library works without passing anything in if those variables are set...is this true of the others? I suppose this is more explicit.  LGTM!

PR #752, opened 2019-03-27
DEVED-3719: authentication with api key and api secret

patillacode:
All snippets have been tested adding the necessary code to send a message to make sure they work.


patillacode:
Thanks @dprothero !

Added now!

PR #769, opened 2019-05-30
Updated deprecated code in sms Twiml generation

GeeH:
I need to update many more snippets as it seems that `Twilio\Twiml` is deprecated in favour of creating specific handlers. I'm not 100% that this is the new way to use the library but once this is confirmed and merged I can update more snippets.


dprothero:
Once we confirm this is the right way, we should update [this project](https://github.com/TwilioDevEd/twiml-generator) as it generates all these PHP code samples based on the input TwiML.

dprothero:
@GeeH We looked into it and this sample is _not_ a generated sample, so we will accept your PR and close it. I've learned the generated samples live under https://github.com/TwilioDevEd/api-snippets/tree/master/twiml and should already be correct with regards to the accepted method of TwiML generation in PHP.

If you see any samples _outside_ of that path that need to be fixed, please do submit an additional PR. Thanks!

PR #784, opened 2019-08-01
Point to the officially supported validation code

eeeschwartz:
To help developers avoid copying this into their code unnecessarily.


eeeschwartz:
I see the python example has been removed so is PR is moot.

PR #785, opened 2019-08-07
Update the 'interpretAs' parameter name to match the ruby helper lib

childish-sambino:


childish-sambino:
Fixes https://github.com/twilio/twilio-ruby/issues/447
Relates to https://github.com/twilio/twilio-ruby/pull/458

PR #803, opened 2019-09-16
[DEVED-4075] Update Reachability code sample

micaswyers:
This commit updates the Reachability indicator ("Process a UserInfo Update") code sample to listen for the `updated` event on the User, rather than the `userUpdated` event on the chatClient.

### DIFF
--- a/ip-messaging/reachability-indicator/process-user-updates/meta.json
+++ b/ip-messaging/reachability-indicator/process-user-updates/meta.json
@@ -1,5 +1,5 @@
 {
-  "title": "Process a UserInfo Update",
-  "description": "Process a UserInfo Update and process the Reachability Indicators",
+  "title": "Process an UpdateReason",
+  "description": "Handle an UpdateReason change and process the Reachability Indicators",
   "type": "server"
 }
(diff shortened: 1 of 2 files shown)


ip-messaging/reachability-indicator/process-user-updates/meta.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Process a UserInfo Update",
-  "description": "Process a UserInfo Update and process the Reachability Indicators",
+  "title": "Process an UpdateReason",
+  "description": "Handle an UpdateReason change and process the Reachability Indicators",
```
micaswyers: ❓ I changed these strings because I found them while grep-ing for the file. What does a meta.json file do in the code sample process?

jefflinwood: I'm pretty sure that that they control the title displayed in the CMS in the header on the code sidebar. You can also see them when you pick an API Snippet to use in a page.

jefflinwood:
This looks good to me!

PR #821, opened 2020-03-20
fix: correct the Ruby client capability token examples

childish-sambino:
Relates to https://github.com/twilio/twilio-ruby/issues/501

### DIFF
--- a/client/capability-token-2way/capability-token.5.x.rb
+++ b/client/capability-token-2way/capability-token.5.x.rb
@@ -4,13 +4,13 @@ require 'sinatra'
 get '/token' do
   account_sid = 'ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'
   auth_token = 'your_auth_token'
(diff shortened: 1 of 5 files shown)


client/capability-token-2way/capability-token.5.x.rb
```diff
@@ -4,13 +4,13 @@
 get '/token' do
   account_sid = 'ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'
   auth_token = 'your_auth_token'
-  capability = Twilio::JWT::Capability.new(account_sid, auth_token)
+  capability = Twilio::JWT::ClientCapability.new(account_sid, auth_token)
```
craigsdennis: AccessToken is the preferred way, but if this fixes the brokenness then 👍

childish-sambino: Yup, mainly about being broken.

childish-sambino:
@craigsdennis Could you merge this?

PR #845, opened 2020-10-13
chore: add Java 8.x snippet which replaces joda.time with java.time

thinkingserious:


eshanholtz:
Is this the only snippet that has a pre 8.x dependency? No examples with guava Range?

thinkingserious:
It's not! Updating now, thank you!

micaswyers (approved):

PR #848, opened 2020-10-17
[DEVEDSB-22] Replace SID placeholders with env vars

mrcrow85:
This addresses [DEVEDSB-22](https://issues.corp.twilio.com/browse/DEVEDSB-22) and [SECRISK-757](https://issues.corp.twilio.com/browse/SECRISK-757).

The aim of this is huge change is to replace all placeholders for sensible account information (such as account sid, auth token, etc.) with environmental variables in order to reduce the risk of users pushing their own credentials on public repos.

The only language missing is curl, which will be done soon.


maylonpedroso (approved):

micaswyers (approved):
So many changes!

PR #869, opened 2021-03-03
Add <Number url> TwiML example.

sarahcstringer:
Adds an example of using the url attribute in the <Number> noun, for [Jira DEVED-4684](https://issues.corp.twilio.com/browse/DEVED-4684)


craigsdennis (approved):
LGTM!

PR #887, opened 2021-04-14
Fix node example so pause is on a separate line.

sarahcstringer:
Chaining pause after say results in an error (`twiml.say(...).pause is not a function`). https://issues.corp.twilio.com/browse/DEVED-5199


stern-shawn (approved):
🎉

PR #895, opened 2021-05-15
[VIDEO-5254] Fix incorrect curl URL in compositions example.

sarahcstringer:
Fix the incorrect URL in the curl example for get-composition-media-file.


ayyrickay (approved):
👍

PR #902, opened 2021-06-07
DEVED-5693: Fix incorrect example for nodejs getting composition media.

sarahcstringer:
Jira: https://issues.corp.twilio.com/browse/DEVED-5693

https://www.twilio.com/docs/video/api/compositions-resource#get-composition-media-file-http-get
Using `response.data` returns undefined. This should be accessing `response.body`.


onlywade (approved):
👍

PR #913, opened 2021-07-12
Correct the SID prefix for Notification resources in sample output

onlywade:
These SIDs were incorrectly prefixed as NO instead of NT.


jefflinwood (approved):
These prefixes look correct

PR #914, opened 2021-07-12
Update client identifier to use underscores instead of hyphens

bld010:
As per the warning on [this page:](https://www.twilio.com/docs/voice/twiml/client) "The client identifier currently may only contain alpha-numeric and underscore characters."

This PR updates the snippets from 'user-jane' to 'user_jane' to reduce confusion for the user.


sarahcstringer (approved):
lgtm

PR #921, opened 2021-07-14
Update Usage Record code samples for consistency

dphamb:
This change adds a print statement to the end of Python examples.


onlywade (approved):
👍

PR #935, opened 2021-08-04
DEVED-5467: iOS Push Snippets - Fix tab spacing, two typos

jefflinwood:


conversations/push-ios/remotenotifications/remotenotifications.m
```diff
@@ -1,6 +1,6 @@
 - (void)application:(UIApplication*)application didRegisterForRemoteNotificationsWithDeviceToken:(NSData*)deviceToken {
     if (self.conversationsClient && self.conversationsClient.user) {
-        [self.TwilioConversationsClient registerWithNotificationToken:deviceToken
+        [self.conversationsClient registerWithNotificationToken:deviceToken
```
anabenites, line 3: A question here, are we not using anymore TwilioConversationsClient Module? or in this case we just needed to fix it and call the function instead? thanks!

anabenites (approved):
It looks good to me, I just left a question to better understand that change.

PR #937, opened 2021-08-11
Sync examples

deshartman:
Propose the below code snippet to add to the docs here: https://www.twilio.com/docs/iam/access-tokens#

I'll also submit the WagTail request once code has been merged

### DIFF
--- a/rest/access-tokens/sync-example/sync-example.5.x.rb
+++ b/rest/access-tokens/sync-example/sync-example.5.x.rb
@@ -1,0 +1,30 @@
+require 'twilio-ruby'
+
+# Required for any Twilio Access Token
+# To set up environmental variables, see http://twil.io/secure
+account_sid = ENV['TWILIO_ACCOUNT_SID']
+api_key = ENV['TWILIO_API_KEY']
+api_secret = ENV['TWILIO_API_KEY_SECRET']
+
+# Required for Sync
+outgoing_application_sid = 'APxxxxxxxxxxxx'
+identity = 'user'
+
+# Create Sync grant for our token
+grant = Twilio::JWT::AccessToken::SyncGrant.new
+grant.outgoing_application_sid = outgoing_application_sid
+
+# Optional: add to allow incoming calls
+grant.incoming_allow = true
+
+# Create an Access Token
+token = Twilio::JWT::AccessToken.new(
+  account_sid,
+  api_key,
+  api_secret,
+  [grant],
+  identity: identity
+)
+
+# Generate the token
+puts token.to_jwt
--- a/rest/access-tokens/sync-example/sync-example.5.x.rb
+++ b/rest/access-tokens/sync-example/sync-example.5.x.rb
@@ -5,18 +5,17 @@ require 'twilio-ruby'
 account_sid = ENV['TWILIO_ACCOUNT_SID']
 api_key = ENV['TWILIO_API_KEY']
 api_secret = ENV['TWILIO_API_KEY_SECRET']
-twilioSyncService = ENV['TWILIO_SYNC_SERVICE_SID']
+twilio_sync_service = ENV['TWILIO_SYNC_SERVICE_SID']
 
 
 # Required for Sync
-outgoing_application_sid = 'APxxxxxxxxxxxx'
 identity = 'user'
 
 # Create Sync grant for our token
 grant = Twilio::JWT::AccessToken::SyncGrant.new
 
 # Create a "grant" which enables a client to use Sync as a given user
-grant.serviceSid = twilioSyncService
+grant.service_sid = twilio_sync_service
 
 # Create an Access Token
 token = Twilio::JWT::AccessToken.new(
--- a/rest/access-tokens/sync-example/sync-example.5.x.rb
+++ b/rest/access-tokens/sync-example/sync-example.5.x.rb
@@ -12,7 +12,6 @@ identity = 'user'
 
 # Create Sync grant for our token
 grant = Twilio::JWT::AccessToken::SyncGrant.new
-grant.outgoing_application_sid = outgoing_application_sid
 
 # Optional: add to allow incoming calls
 grant.incoming_allow = true
--- a/rest/access-tokens/sync-example/sync-example.5.x.rb
+++ b/rest/access-tokens/sync-example/sync-example.5.x.rb
@@ -5,6 +5,8 @@ require 'twilio-ruby'
 account_sid = ENV['TWILIO_ACCOUNT_SID']
 api_key = ENV['TWILIO_API_KEY']
 api_secret = ENV['TWILIO_API_KEY_SECRET']
+twilioSyncService = ENV['TWILIO_SYNC_SERVICE_SID']
+
 
 # Required for Sync
 outgoing_application_sid = 'APxxxxxxxxxxxx'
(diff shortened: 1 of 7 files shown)


deshartman:
Just realised I left "outgoingApplicationSid" in. Have updated it, but unsure how to update PR to the latest commit

bld010:
Thanks for adding this. It definitely seems necessary and helpful to get all of the Grants examples in one place!

I'm not super-familiar with Sync yet, but I notice that [the JavaScript example that exists in docs right now](https://www.twilio.com/docs/sync/identity-and-access-tokens#minting-an-access-token-in-your-backend) has a lot of information about `identity`, and the identity is both attached to the token AND sent as part of the token response from the /token endpoint. 

Specifically, it has these comments: 
```js  
 
    // This is the most critical part of your backend code, as you must identify the user and (possibly)
    // challenge them with some authentication scheme. To determine the identity, you might use:
    //    * A session datum consistently identifying one anonymous visitor,
    //    * A session key identifying a logged-in user
    //    * OAuth credentials identifying a logged-in user
    //    * A random username for all comers.
    //
```
and we see it added to the token and sent in the response separate from the token itself: 

```js
token.identity = identity;

    // Serialize the token to a JWT string and include it in a JSON response
    response.send({
        identity: identity,
        token: token.toJwt()
    });
```

@anabenites Do you have any insight into whether we need to include these comments in the SyncGrant examples for [this page](https://www.twilio.com/docs/iam/access-tokens)? Is there more detailed information about how to authenticate a user's identity, or is this just a "you can do this if necessary" kind of thing? 

Also, @anabenites the SyncGrant can also take an `endpointId` but I couldn't find any information on that. Do we need to include that in these examples? Thanks!

anabenites:
hi @bld010 I asked and the `endpointId` is deprecated. I think there is no need to add the additional comments because the users should build what best suit their needs. Are we still using `outgoingApplicationSid`? In my case is throwing an error. I was also thinking that to create a SyncGrant we need to identify the Sync Service instance something like this:
```
const syncGrant = new SyncGrant({
  serviceSid: serviceSid,
  incomingAllow: true, // Optional: add to allow incoming calls
});
```

deshartman:
@anabenites "outgoingApplicationSid" was a mistake. If you look, I did an additional commit after the PR to remove it. My mistake

anabenites:
@deshartman ok, thanks for the information. A question why are you passing `incomingAllow: true, // Optional: add to allow incoming calls` to new SyncGrant? thanks,

deshartman:
@anabenites I just realised, I checked in the wrong code. I had meant to delete all the voice related grants and only add the Sync ones. My mistake. Sorry about this. Have not done PRs before, so still learning.

I have completely reworked the examples, and re-uploaded. Hope my other language proposals are ok?

rest/access-tokens/sync-example/sync-example.5.x.rb
```diff
@@ -0,0 +1,31 @@
...
+# Required for any Twilio Access Token
+# To set up environmental variables, see http://twil.io/secure
+account_sid = ENV['TWILIO_ACCOUNT_SID']
+api_key = ENV['TWILIO_API_KEY']
+api_secret = ENV['TWILIO_API_KEY_SECRET']
+twilioSyncService = ENV['TWILIO_SYNC_SERVICE_SID']
+
+
+# Required for Sync
+outgoing_application_sid = 'APxxxxxxxxxxxx'
```
anabenites, line 12: Hi, I think that we need to remove here also `outgoing_application_sid`.

rest/access-tokens/sync-example/sync-example.6.x.py
```diff
@@ -0,0 +1,27 @@
+import os
+from twilio.jwt.access_token import AccessToken
+from twilio.jwt.access_token.grants import SyncGrant
+
+# required for all twilio access tokens
+# To set up environmental variables, see http://twil.io/secure
+account_sid = os.environ['TWILIO_ACCOUNT_SID']
+api_key = os.environ['TWILIO_API_KEY']
+api_secret = os.environ['TWILIO_API_KEY_SECRET']
+twilioSyncService = os.environ[process.env.['TWILIO_SYNC_SERVICE_SID']
```
anabenites, line 10: Here, instead of `twilioSyncService` you can use `twilio_sync_service` because the name convention in Python is snake_case.

rest/access-tokens/sync-example/sync-example.6.x.py
```diff
@@ -0,0 +1,27 @@
...
+# required for Sync grant
+identity = 'user'
+
+# Create access token with credentials
+token = AccessToken(account_sid, api_key, api_secret, identity=identity)
+
+# Create a Sync grant and add to token
+sync_grant = SyncGrant(
+    # Create a "grant" which enables a client to use Sync as a given user
+    serviceSid=twilioSyncService
```
anabenites, line 21: here the same as above, you can write `service_sid=twilio_sync_service`

rest/access-tokens/sync-example/sync-example.5.x.rb
anabenites, line 12: We can remove `outgoing_application_sid` because we are not using it anymore.

rest/access-tokens/sync-example/sync-example.5.x.rb
```diff
@@ -0,0 +1,31 @@
+require 'twilio-ruby'
+
+# Required for any Twilio Access Token
+# To set up environmental variables, see http://twil.io/secure
+account_sid = ENV['TWILIO_ACCOUNT_SID']
+api_key = ENV['TWILIO_API_KEY']
+api_secret = ENV['TWILIO_API_KEY_SECRET']
+twilioSyncService = ENV['TWILIO_SYNC_SERVICE_SID']
```
anabenites, line 8: we are using for the variables also snake_case convention, let's use it also for `twilio_sync_service`.

rest/access-tokens/sync-example/sync-example.5.x.rb
```diff
@@ -0,0 +1,31 @@
...
+
+# Required for Sync
+outgoing_application_sid = 'APxxxxxxxxxxxx'
+identity = 'user'
+
+# Create Sync grant for our token
+grant = Twilio::JWT::AccessToken::SyncGrant.new
+
+# Create a "grant" which enables a client to use Sync as a given user
+grant.serviceSid = twilioSyncService
```
anabenites, line 19: the name of the argument is `service_sid`. We can update also `twilio_sync_service` here.

deshartman:
@anabenites Changes done and pushed. Good suggestions!

rest/access-tokens/sync-example/sync-example.6.x.py
```diff
@@ -0,0 +1,27 @@
...
+# required for Sync grant
+identity = 'user'
+
+# Create access token with credentials
+token = AccessToken(account_sid, api_key, api_secret, identity=identity)
+
+# Create a Sync grant and add to token
+sync_grant = SyncGrant(
+    # Create a "grant" which enables a client to use Sync as a given user
+    serviceSid=twilio_sync_service
```
anabenites, line 21: The correct argument is `service_sid` , so it will be `service_sid=twilio_sync_service` 
For reference: https://github.com/TwilioDevEd/sdk-starter-python/blob/master/app.py#L90

anabenites (approved):
thanks for the changes, it looks good to me!

PR #940, opened 2021-09-22
remove wrong call to 'accounts' method

jalbertsr:
I was using node.js and it was not working for me, then I checked how it was in other languages and I saw this was wrongly being used there. I guess the account SID is already set in the `client` instantiation.

### DIFF
--- a/rest/messages/feedback-confirm/feedback-confirm.3.x.js
+++ b/rest/messages/feedback-confirm/feedback-confirm.3.x.js
@@ -19,7 +19,6 @@ app.get('/confirm', (req, res) => {
 
   // Send Feedback to Twilio
   client
-    .accounts(accountSid)
     .messages(messageSid)
     .feedback.create({
       outcome: 'confirmed',


jalbertsr:
@stern-shawn can you check it pls, does it makes sense or it need some changes?

stern-shawn (approved):
Yep, this looks good to me, thanks for the catch!

Working in local dev w/ latest v3 Twilio:
Screen Shot 2021-09-23 at 11 32 24 AM

I don't think accounts is even a valid method anymore...
image

stern-shawn:
For a bit of context, which page did you find this broken example on? 🙇‍♂️

jalbertsr:
Yes, you're right [this](https://www.twilio.com/docs/sms/send-message-feedback-to-twilio?code-sample=code-send-delivery-confirmation-feedback&code-language=Node.js&code-sdk-version=3.x) here is where I found it :)

Issue #950, opened 2021-11-24
PHP 6.x Snippets


dooley1001:
There is no snippets available for the newer version of PHP - As seen here, current version is outdated - https://www.twilio.com/docs/usage/security

PR #951, opened 2021-11-24
feat: update signature_validation php sample from 5.x to 6.x

stern-shawn:
Addresses #950

Appears to be working on my machine, just need to update the file extension to `6.x`. (Or, is it best practice to keep the old file and create a new, separate `6.x` file?)

```bash
❯ php security.php
Confirmed to have come from Twilio.%
```


sarahcstringer (approved):
🙌

PR #959, opened 2021-12-30
fix: standardize Runtime mobile sample code

stern-shawn:
Part of our revamp of the Runtime docs has included consistently using ES6 arrow functions as well as making sure users get in the habit of using `return callback` instead of just `callback`. It's not technically an issue for these examples, but it helps with the consistency of our samples and messaging :)


jefflinwood (approved):
Approved!

PR #961, opened 2022-01-24
replace var with const for variable declaration

stuyy:
Update variable declarations: Remove var and replaced with const, for this document: https://www.twilio.com/docs/taskrouter/js-sdk/workspace/worker


dfokina (approved):

stern-shawn (approved):
Thanks! No reason for `var` in 2022 🚢 🇮🇹

PR #964, opened 2022-03-01
feat: update serverless cookie example from 4->24 hours

stern-shawn:


stern-shawn:
Turns out this doesn't work. Closing.

PR #977, opened 2022-06-29
Add examples for <Connect><Conversation> TwiML

bld010:
I created snippets for hopefully every attribute available for <Connect><Conversation> use. (Why so many? Because our Java and C# friends' libraries don't use strings for most attribute values. This means they would either need to guess, rely on autocomplete in their IDEs, or look at the autogenerated documentation for the libraries.)

Our twiml-generator tool can't always spit out the right enums/lists, so I had to manually change and verify C# and Java examples. I'd love some extra 👀  around those files. I have a strong feeling some of my implementation choices are wrong or at least not standard. 

Looking for eyes on: 
- Naming files correctly/structuring files correctly
- Java/C# Enums/Lists usage
- Ruby/Python/PHP review? (The twiml-generator tool doesn't complain much about these, so just want to make sure they look correct)


twiml/voice/connect/conversation/conversation-action-method/conversation-action-method.3.x.js
```diff
@@ -0,0 +1,12 @@
+const VoiceResponse = require('twilio').twiml.VoiceResponse;
+
+const response = new VoiceResponse();
+const connect = response.connect({
+    action: 'https://your-action-url.com/',
```
sarahcstringer, line 5: You can take or leave this suggestion -- I've started using example.com for all my fake URLs, because that's a [dedicated URL to be used for demo purposes ](https://datatracker.ietf.org/doc/html/rfc2606#section-3)(as opposed to trying to come up with a URL that doesn't exist to demo). It's just a personal preference that I haven't actually brought up with the team, though.

twiml/voice/connect/conversation/conversation-action-method/meta.json
```diff
@@ -0,0 +1,5 @@
+{
+    "title": "Use <Connect>'s action and method attributes with <Conversation>",
+    "type": "server",
+    "test_output": true
+}
```
sarahcstringer, line 5: Small nit -- looks like some of the files have linebreaks at the end and some don't.

twiml/voice/connect/conversation/conversation-action-method/conversation-action-method.7.x.py
```diff
@@ -0,0 +1,8 @@
+from twilio.twiml.voice_response import Connect, Conversation, VoiceResponse
```
sarahcstringer, line 1: Nit: I think this is something wrong with our twiml-generator too, but it unnecessarily imports the `Conversation` object. You could get rid of the `Conversation` imports at the top level, because they aren't used anywhere. It doesn't break things to keep it in, it's just not used.

sarahcstringer (approved):
This LGTM. I left a few nits that you can take or leave. I'm not a Java/C# expert, but from some quick looking, it seems like you're using the right methods for creating lists and using enums.

twiml/voice/connect/conversation/conversation-action-method/meta.json
bld010: Yes, it seems like its the files I manually created 🤦🏻 haha. Thanks!

twiml/voice/connect/conversation/conversation-action-method/conversation-action-method.3.x.js
bld010: Yes, good point! I just want to avoid just `http://www.example.com/` because sometimes the lack of context of what the URL should be is confusing ... but I can do `example.com/your-action-handler` or something instead!

bld010:
@sarahcstringer I've added example.com, newlines, a missing set of examples, and a missing meta.json file.

sarahcstringer (approved):

PR #981, opened 2022-07-06
Update file extensions to current versions

antonhi:
https://issues.corp.twilio.com/browse/DEVED-7458


sarahcstringer:
I went through the files and it looks like they were previously updated incorrectly :-\ so they don't work currently.

- The credentials aren't correct in the examples (actually for all of them in this folder, not just the ones you changed...). The file uses API key and secret, but there's no account SID, which you'd also need to include to get this to work.
- Instead of using an API key and secret, I'd just go with account sid and auth token, like all our other dynamic code samples use. You can just copy/paste an existing example's client setup for each language, such as [this one](https://www.twilio.com/docs/video/api/compositions-resource?code-sample=code-transcode-a-video-recording&code-language=PHP&code-sdk-version=6.x)
- Java: this example doesn't seem to work; I think there are some missing imports. I'll keep digging a bit more.

video/rest/compositions/get-composition-media-file/get-composition-media-file.5.x.cs
```diff
@@ -5,21 +5,25 @@
 using System.IO;
 using System.Net;
 using System.Text;
+using Twilio;
```
sarahcstringer, line 8: 🤦‍♀️ 
thanks for fixing!!

video/rest/compositions/get-composition-media-file/get-composition-media-file.8.x.java
```diff
@@ -0,0 +1,41 @@
+// Install the Java helper library from twilio.com/docs/java/install
```
sarahcstringer, line 1: When I copy this and run it, I get a few errors about some weird special characters:
```./Main.java:17: error: illegal character: '\u200b'

^
./Main.java:23: error: illegal character: '\u200b'

^
./Main.java:31: error: illegal character: '\u200b'

^
./Main.java:33: error: illegal character: '\u200b'

^
./Main.java:37: error: illegal character: '\u200b'
```

I guess somehow some empty whitespace characters popped into the file?

Here's an example command you can run to get rid of them: https://superuser.com/questions/207207/how-can-i-delete-u200b-zero-width-space-using-sed

Or a suggestion about how to do it with VSCode https://stackoverflow.com/questions/32013346/zero-width-space-character-in-string-literals-detect-and-or-prevent

sarahcstringer: I think this file is also missing some of the required imports. The one you sent me in Slack works though (after removing the special characters).

video/rest/compositions/get-composition-media-file/get-composition-media-file.5.x.cs
```diff
@@ -5,21 +5,25 @@
...
+      string authToken = Environment.GetEnvironmentVariable("TWILIO_AUTH_TOKEN");
+
+      TwilioClient.Init(accountSid, authToken);
 
       const string compositionSid = "CJXXXX";
       const string uri = $"https://video.twilio.com/v1/Compositions/{compositionSid}/Media?Ttl=3600";
 
       var request = (HttpWebRequest)WebRequest.Create(uri);
-      equest.Headers.Add("Authorization", "Basic " + Convert.ToBase64String(Encoding.ASCII.GetBytes(apiKeySid + ":" + apiKeySecret)));
+      request.Headers.Add("Authorization", "Basic " + Convert.ToBase64String(Encoding.ASCII.GetBytes(apiKeySid + ":" + apiKeySecret)));
```
sarahcstringer, line 26: After changing the `apiKeySid` and `apiKeySecret` variables here to be the account sid and auth token, this *mostly* works. It doesn't download an mp4, but instead downloads ... I'm not exactly sure what. Maybe the binary data? I'll look more tomorrow :)

sarahcstringer: But, it runs and gets the file from Twilio

sarahcstringer:
Thanks for these updates, it's looking good!
- The PHP, Ruby, and Python ones all work 🎉 
- The Java and C# ones are looking close
- The Node one technically works but I think we should update that one to use Auth Tokens rather than API keys, just to fit with all the other samples here.

video/rest/compositions/get-composition-media-file/get-composition-media-file.3.x.js
```diff
@@ -2,18 +2,17 @@
 // information on how to download and install this version, visit
 // https://www.twilio.com/docs/libraries/node
```
sarahcstringer, line 3: One more nit, this header section can go away because we cover it a bit later.

sarahcstringer (approved):
This looks great!! Thanks for all your work on it :shipit:

PR #985, opened 2022-07-28
Add example for subscribing a Media Transcriber to a single Room Participant.

sarahcstringer:
This essentially copies the code from [an existing code sample](https://github.com/TwilioDevEd/api-snippets/tree/master/video/rest/rooms/participants/update-subscribe-rules-static) and only changes the specific rule being applied.


antonhi (approved):
Looks good!

PR #988, opened 2022-09-01
Tidy messaging express

stern-shawn:
Getting ready to write Golang versions of these, and realized the Express samples are using some janky methods / involve more typing than necessary 👀


bld010 (approved):
LGTM!

PR #993, opened 2022-09-23
Add Dialogflow CX VirtualAgent examples.

sarahcstringer:
Adds code samples for <Param> and <Config> nouns within <VirtualAgent>.

The Node.js example doesn't currently work because of a bug in the helper library, but I've tested the other examples.


bld010 (approved):
LGTM! One question ... `uniqueName` to me implies that the name must not include spaces. Is that the case?

PR #997, opened 2022-12-01
Fix compiler error on line 20

Aaronstotle:
This snippet gives a compiler warning that you cannot assign an error type to a string, err.Error() casts it correcty and allows it to compile. 

Original code gives this compiler error: cannot use err (variable of type error) as string value in argument to context.String


Aaronstotle:
Context: I was using this code to follow along the getting started with Twilio and golang and ran into this error after copying and pasting. Figured it would be helpful to fix this for other folks in the future 😄

stern-shawn (approved):
I appreciate the catch, thank you for looking out for others!

PR #998, opened 2022-12-09
DEVED-7951: Add Identity Match Mobile Identity API code samples

amason-twilio:


mobile-identity/identity-match/identity-match-example.3.x.js
```diff
@@ -0,0 +1,28 @@
...
+  postalCode: '51501',
+  countryCode: 'US',
+  nationalId: '345678901',
+  dateOfBirth: '19800303'
+}, {
+  authorization: 'y5jBqXWLzEF4OgK61EbJ4jbEesmxE5No',
+  requesttime: '2022-08-01T09:33:43+00:00'
+})
+  .then(({ data }) => console.log(data))
+  .catch(err => console.error(err));
```
sarahcstringer, line 28: Nit: these should have linebreaks at the end of each file

sarahcstringer:
Have you tested all these?

mobile-identity/identity-match/identity-match-example.5.x.cs
```diff
@@ -0,0 +1,12 @@
+//Installation
```
sarahcstringer, line 1: I wonder if there's a linter you could run these through? Maybe this is fine -- I don't know about csharp formatting -- but it seems like a big block that could maybe be broken up.

mobile-identity/identity-match/identity-match-example.3.x.js
```diff
@@ -0,0 +1,28 @@
...
+  address1: '3701 Ashwood Dr',
+  address2: 'Apt 20',
+  city: 'Council Bluffs',
+  state: 'IA',
+  postalCode: '51501',
+  countryCode: 'US',
+  nationalId: '345678901',
+  dateOfBirth: '19800303'
+}, {
+  authorization: 'y5jBqXWLzEF4OgK61EbJ4jbEesmxE5No',
```
sarahcstringer, line 24: What's this? Should this be stubbed or anything?

mobile-identity/identity-match/identity-match-example.3.x.js
amason-twilio: Updated to include linebreaks, thanks!

mobile-identity/identity-match/identity-match-example.3.x.js
amason-twilio: This is an example security key used to uniquely access this API. This API is a weird one and doesn't use the Twilio API Base URL or Twilio Auth for access. 

Do you think I should replace with something like "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"?

mobile-identity/identity-match/identity-match-example.5.x.cs
amason-twilio: I also don't know haha - I will look into this!

amason-twilio:
@sarahcstringer 

> Have you tested all these?

I haven't since they were pre-existing on this page https://mobileidentity.twilio.com/reference/post_v2-match. Given that, do you recommend I test? Also curious if you mean manually test or writing tests for these.

mobile-identity/identity-match/identity-match-example.3.x.js
```diff
@@ -0,0 +1,28 @@
+// Installation
+// $ npm install api --save
```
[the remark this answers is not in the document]
amason-twilio: Resolving per this Slack thread: https://twilio.slack.com/archives/GMDHTPDGV/p1670608168874619

mobile-identity/identity-match/identity-match-example.5.x.cs
amason-twilio: Updated so each param is on its own line.

sarahcstringer (approved):

PR #1007, opened 2023-02-07
docs: updated csharp snippets version to 6

sbansla:
updated csharp snippets version to 6


stern-shawn (changes requested):
Please address merge conflicts and then we'll be good to approve :)

PR #1011, opened 2023-03-02
fix: ruby prosody argument for words

stern-shawn:
Addresses an issue in the sample code for ruby + prosody: https://github.com/twilio/twilio-ruby/issues/637

edit: this should be future-proof at least for the upcoming release, since the (admittedly, terrible) [argument setup](https://github.com/twilio/twilio-ruby/blob/6.0.0-rc/lib/twilio-ruby/twiml/voice_response.rb#L596) is maintained 🙃


bld010 (approved):
LGTM ... thank you!

PR #1014, opened 2023-03-14
docs: updated csharp snippets version to 6

sbansla:
twilio-csharp is on version 6, thus upgrading snippets versions.


stern-shawn (changes requested):
If you could also update the version numbers for csharp in README.md, that'd be fantastic :) 

https://github.com/TwilioDevEd/api-snippets#:~:text=snippet%2Dexample%2D1.4.x.cs%0Asnippet%2Dexample%2D1.5.x.cs
ex.
```
snippet-example-1.5.x.cs
snippet-example-1.6.x.cs
```

stern-shawn (approved):

PR #1022, opened 2023-04-27
Say snippets

bld010:
Add new `<Say>` snippets to match rewrite of `<Say>` docs
Removed the TTS markdown files (because they didn't work with Wagtail)


twiml/voice/say/say-basic-usage/say-basic-usage.8.x.py
```diff
@@ -0,0 +1,6 @@
+from twilio.twiml.voice_response import VoiceResponse, Say
```
sarahcstringer, line 1: There's a small bug with the TwiML generator where it imports all the TwiML classes technically used, even though this one only needs VoiceResponse (because the `.say` is added to VoiceResponse, and it doesn't actually need to import the `Say` class). It does this for everything. It's just an unnecessary import if you wanted to update it, but no worries if not :)

sarahcstringer (approved):

twiml/voice/say/say-basic-usage/say-basic-usage.8.x.py
bld010: Ah interesting. Thanks for the info! I can add that to the twiml-generator repo's issues.

PR #1026, opened 2023-05-12
Fix TwiML snippets for Messaging Redirect.

sarahcstringer:
https://issues.corp.twilio.com/browse/DEVED-7129

The TwiML examples for the Messaging `<Redirect>` verb were using `VoiceResponse`s instead of `MessagingResponse`s.

I've run the code snippets through the TwiML generator verify tool.

Small change: changed the sample URL from foo.com to example.com.


bld010 (approved):
🚢

PR #1037, opened 2023-07-26
Move code snippets from docs to GitHub

sarahcstringer:
https://issues.corp.twilio.com/browse/DEVED-8517

Takes the code samples from this page https://www.twilio.com/docs/global-infrastructure/create-an-outbound-call-via-rest-api-in-a-non-us-twilio-region and puts them into this repo.

The Tabbed Content component in Wagtail only supports up to 6 code samples, so I can't add another one for Go in the existing doc as it is.

I tested the new Go code sample, and the others are taken directly from the docs.

### DIFF
--- a/voice/specify-edge/meta.json
+++ b/voice/specify-edge/meta.json
@@ -1,0 +1,4 @@
+{
+  "title": "Specify a Twilio Region and Edge when making a call",
+  "type": "server"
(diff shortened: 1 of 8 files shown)


thinkingserious (approved):
LGTM 👍🏼

PR #1047, opened 2024-02-07
chore: changed file names to 10.x

sbansla:
As a part of MVR Release:
1. Changed file names to 10.x
2. Changed version from Readme.md file


tomwanzek (approved):
🚀 As per https://github.com/twilio/twilio-java/blob/main/UPGRADE.md there are no breaking changes in the Java Helper Library v10 release, so there should be no need to change API Snippet content.

PR #1057, opened 2024-04-29
Update link-shortening-domain-cert.10.x.java

vlim-twilio:
Fixing typo ('Sring' to 'String') on line 12


bermudezmt (approved):
Thanks for the fix!

PR #1078, opened 2024-06-12
Benparj98 content sdk samples

benparj98:
adding most C# and Java sdk samples to content create


bermudezmt (approved):
This has been technically validated by Cedric Barbez and Justin Helgerson.

PR #1081, opened 2024-07-10
Add Transcription TwiML snippets

bld010:
TwiML snippets that will replace the static code samples on https://www.twilio.com/docs/voice/twiml/transcription

These snippets have been tested against the most recent version of the Helper Libraries -- the code does output the associated TwiML. 

Golang is missing, this is a known issue. A separate ticket will be created, since these need to be written manually. The twiml-generator project doesn't have Golang support.

### DIFF
--- a/twiml/voice/transcription/transcription-basic-usage/meta.json
+++ b/twiml/voice/transcription/transcription-basic-usage/meta.json
@@ -1,0 +1,4 @@
+{
+    "title": "<Transcription> - basic usage",
+    "type": "server"
+}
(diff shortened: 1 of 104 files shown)


bld010:
To check this PR: 

- Make sure that each snippet directory has: 
  - an /output/ directory with a .twiml file 
  - six code samples (.cs, .php, .java, .js, .rb, .php)
  - a meta.json file that matches the associated code samples in that directory

sarahcstringer (approved):

